PR #9, opened 2022-01-22
Bugfix/P2022 286 fix eslint rules and syntax problems

pawelwojdas:


kacperChlon:
Please change target branch to develop, master should be only used in case of version release etc...

package.json
```diff
@@ -61,13 +61,17 @@
     "@emotion/react": "^11.7.1",
```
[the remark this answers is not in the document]
kacperChlon: Use this: https://thechrisgreen.com/2021/08/vs-code-sort-lines-of-code-in-ascending-or-descending-order/ ;)

.eslintrc.json
```diff
@@ -12,13 +12,16 @@
       "rules": {
         "import/export": 0,
         "import/no-self-import": 0,
+        "global-require": 0,
```
kacperChlon, line 15: Ok, but why? :)

pawelwojdas: I forgot to remove this line :)

FabianHoffmannIntive (approved):

.storybook/tsconfig.json
```diff
@@ -10,5 +10,8 @@
     "../**/*.spec.jsx",
     "../**/*.test.jsx"
   ],
-  "include": ["../**/*"]
+  "include": ["../**/*"],
+  "compilerOptions": {
+    "outDir": ""
```
Adys23, line 15: Out of curiosity, what is the expected result of adding this option? How is it related to eslint and syntax problems? shouldn't it be:
`"outDir": "/dist"`

Adys23 (approved):

.storybook/tsconfig.json
pawelwojdas: My bad, you're definitely right 😄 should be as you said. 
Problem is described here: https://www.codejam.info/2021/10/typescript-cannot-write-file-overwrite-input.html

pawelwojdas: I added this option to remove problem: "cannot write file.... because it would overwrite input file". I followed convention as is in file: "libs/ui/.storybook/tsconfig.json". Dist folder is only for the code used in production. As I understood storybook is only for development. 

More about this problem you can find here -> https://www.codejam.info/2021/10/typescript-cannot-write-file-overwrite-input.html

Adys23: I have read the article after your first comment and started wondering "do we really need strybooks to be compiled to production?" and yeah, I do not see the point in that. Wondering why you were the only one who had this issue and it was working fine without these lines for everyone else...

pawelwojdas: Probably everyone had these problems before. They're not errors, the environment works fine, but just some ESLint problems may occur.

PR #10, opened 2022-01-25
JI:758031 P2022-235 Test task-Aleksandra Smogura

Reden4005:
Add a test component `TestButton` with story and e2e test.

Closes [P2022-235]

### DIFF
--- a/libs/features/feedback/src/components/test-button/index.tsx
+++ b/libs/features/feedback/src/components/test-button/index.tsx
@@ -11,8 +11,4 @@ export interface TestButtonProps {
   title: string;
 }
 
-export const TestButton: React.FC<TestButtonProps> = ({ variant, title }) => {
-  return <Button variant={variant}>{title}</Button>;
-};
-
-export default TestButton;
+export const TestButton: React.FC<TestButtonProps> = ({ variant, title }) => <Button variant={variant}>{title}</Button>;
--- a/libs/features/feedback/src/components/test-button/index.tsx
+++ b/libs/features/feedback/src/components/test-button/index.tsx
@@ -1,7 +1,13 @@
 import { Button } from "@mui/material";
 
+export enum TestButtonVariant {
+  text = "text",
+  outlined = "outlined",
+  contained = "contained"
+}
+
 export interface TestButtonProps {
-  variant: "text" | "outlined" | "contained" | undefined;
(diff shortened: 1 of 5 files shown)


kacperChlon:
Please use tampermonkey script to get proper name for the PR

Reden4005:
> Please use tampermonkey script to get proper name for the PR

I changed it. Is it now ok?

libs/features/feedback/src/components/test-button/index.tsx
```diff
@@ -0,0 +1,12 @@
+import { Button } from "@mui/material";
+
+export interface TestButtonProps {
+  variant: "text" | "outlined" | "contained" | undefined;
```
kacperChlon, line 4: Instead of sample string values preferable way to declare predefined values is: enum type or just type, hence you can change it, moreover you can declare optional parameter by adding ? at the end: https://www.typescriptlang.org/docs/handbook/advanced-types.html#optional-parameters-and-properties

libs/features/feedback/src/components/test-button/index.tsx
```diff
@@ -0,0 +1,12 @@
...
+  title: string;
+}
+
+export const TestButton: React.FC<TestButtonProps> = ({ variant, title }) => {
+  return <Button variant={variant}>{title}</Button>;
```
kacperChlon, line 9: if no code is added to the body of the component please use simplified form without return keyword.

libs/features/feedback/src/components/test-button/index.tsx
```diff
@@ -0,0 +1,12 @@
...
+};
+
+export default TestButton;
```
kacperChlon, line 12: We prefer export const over defaults ;)

libs/features/feedback/src/components/test-button/index.tsx
Reden4005: Please check.

libs/features/feedback/src/components/test-button/index.tsx
Reden4005: removed line

libs/features/feedback/src/components/test-button/index.tsx
Reden4005: Changed

PR #12, opened 2022-01-28
JI:759426 P2022-461 Application routing

Reden4005:
Add basic routes with sample content.


apps/site/src/app/components/index.ts
```diff
@@ -1 +1,2 @@
 export * from ".";
+export * from "./routing";
```
kacperChlon, line 2: I would say routing is more like util not component itself

apps/site/src/app/components/routing/index.tsx
```diff
@@ -0,0 +1,35 @@
+import { Navigate, Outlet, Route, Routes } from "react-router-dom";
+
+export enum ROUTES {
```
kacperChlon, line 3: Enum types should have: 
- Singular form
- PascalCased

since routes names will be used in many places in whole project I recommend to move this to shared types dir

apps/site/src/app/components/routing/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+  presentation = "/presentation",
+  add = "add",
+  edit = "edit/:id",
+  external = ":id",
+  error = "*"
+}
+
+export const Routing: React.FC = () => {
+  return (
+    <Routes>
```
kacperChlon, line 15: Please make a usage of this [hook approach](https://reactrouter.com/docs/en/v6/api#useroutes) which also promotes functional programming ;)

libs/ui-e2e/src/integration/ui/routing/routing.spec.ts
```diff
@@ -0,0 +1,40 @@
+describe("testing basic routing", () => {
```
kacperChlon, line 1: It is a test itself, no need to add it into the desc name. Base routing is sufficient

libs/ui-e2e/src/integration/ui/routing/routing.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+  beforeEach(() => cy.visit("localhost:4000/"));
+
+  it("should redirect and display home page", () => {
+    cy.get("h1").contains("Home page");
```
kacperChlon, line 5: Since all of those will change pretty quickly it will require from you to refactor the tests, hence I recommend to check the path of the page instead of header, which will be removed.

libs/ui-e2e/src/integration/ui/routing/routing.spec.ts
```diff
@@ -0,0 +1,40 @@
+describe("testing basic routing", () => {
+  beforeEach(() => cy.visit("localhost:4000/"));
```
kacperChlon, line 2: The root/base path can be set in cypress.json for given project then you can use sth like cy.visit('/').

libs/ui-e2e/src/integration/ui/routing/routing.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+  });
+
+  it("should display presentation/add page", () => {
+    cy.visit("localhost:4000/presentation/add");
+    cy.get("h1").contains("Presentation page");
+    cy.get("h1").contains("Add presentation page");
+  });
+
+  it("should display presentation/edit page", () => {
+    cy.visit("localhost:4000/presentation/edit/someID");
```
kacperChlon, line 25: Instead of using 'someID' let's use something which looks more like real data, for this purpose let's use library uuid to generate unique identifier for us and use it here.

libs/ui-e2e/src/integration/ui/routing/routing.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+    cy.get("h1").contains("Presentation page");
+    cy.get("h1").contains("Edit presentation page");
+  });
+
+  it("should display presentation/eternal page", () => {
```
kacperChlon, line 30: If you will put routes enum into the shared directory you can import those enum and use in inside the test name like normal variable: 

`should display <proper_path_variable> page`

kacperChlon:
Please move e2e tests into the proper project. 
UI-e2e is for stories from storybooks,
Site-e2e is a project which should contain all tests related with site project (where routing belongs)

apps/site/src/app/components/routing/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+      <Route path='/' element={<Navigate replace to={ROUTES.home} />} />
+      <Route path={ROUTES.home} element={<h1>Home page</h1>} />
```
kacperChlon: What is the difference between 'home' and '/' path?

apps/site/src/app/components/index.ts
Reden4005: Moved to site/src/app/utils/routing;

apps/site/src/app/components/routing/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+            <h1>Presentation page</h1>
+            <Outlet />
+          </>
+        }
+      >
+        <Route path={ROUTES.add} element={<h1>Add presentation page</h1>} />
+        <Route path={ROUTES.edit} element={<h1>Edit presentation page</h1>} />
+        <Route path={ROUTES.external} element={<h1>Page for external user</h1>} />
+      </Route>
+      <Route path={ROUTES.error} element={<h1>You should not be here...</h1>} />
```
kacperChlon: I would rather use name like NotFound in this case since it's not an error, it's just the page we don't support/can't find ;)

libs/ui-e2e/src/integration/ui/routing/routing.spec.ts
Reden4005: I had problem with this, to resolve it I changed "baseUrl": from 4400 to "http://localhost:4000" in libs/ui-e2e. Now it is working.

apps/site/src/app/components/routing/index.tsx
Reden4005: done

libs/ui-e2e/src/integration/ui/routing/routing.spec.ts
Reden4005: Done

apps/site/src/app/components/routing/index.tsx
Reden4005: done

libs/ui-e2e/src/integration/ui/routing/routing.spec.ts
Reden4005: chenged

libs/ui-e2e/src/integration/ui/routing/routing.spec.ts
Reden4005: Done

apps/site/src/app/components/routing/index.tsx
Reden4005: Changed

apps/site/src/app/components/routing/index.tsx
Reden4005: Changed Route path to "/" for Home page.

libs/ui-e2e/src/integration/ui/routing/routing.spec.ts
Reden4005: Done

apps/site/src/app/index.ts
```diff
@@ -1,2 +1,4 @@
 export * from "./components";
 export * from "./pages";
+export * from "./utils/appLink";
+export * from "./utils/routing";
```
Adys23, line 4: To keep the export convention here we should export all from utils folder:
`export * from "./utils"`
and utils folder should have it's own index.ts with:
`export * from "./appLink"`
`export * from "./routing"`

libs/shared/src/index.ts
```diff
@@ -1,5 +1,6 @@
 export * from "./components";
 export * from "./hooks";
 export * from "./types";
+export * from "./types/appRoutes";
```
Adys23, line 4: You can export appRoutes directly in "types/index.ts" same as it is done for translations.ts

Adys23 (commented):
Same as in my case :) 
Folders and files should have names with lowercase (kebab case) (https://betterprogramming.pub/string-case-styles-camel-pascal-snake-and-kebab-case-981407998841)

apps/site/src/app/components/index.ts
```diff
@@ -1 +1,2 @@
+export * from "../utils/routing";
```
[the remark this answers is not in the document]
Reden4005: Right, deleted.

apps/site/src/app/index.ts
Reden4005: Changed

libs/shared/src/index.ts
Reden4005: Done.

Adys23 (approved):

apps/site/src/app/utils/appLink/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+  const { to, lang, children, linkId, ...rest } = props;
+  let path;
+
+  if (to === AppRoute.EditPresentation || to === AppRoute.PresentationForExternalUser || to === AppRoute.AddPresentation) {
+    path = `${generatePath(`${AppRoute.Presentation}/${to}`, { id: linkId })}?lang=${lang}`;
+  } else {
+    path = `${generatePath(`${to}`)}?lang=${lang}`;
+  }
+  return (
+    // eslint-disable-next-line react/jsx-props-no-spreading
```
FabianHoffmannIntive, line 20: We should not turn-off eslint rules

apps/site/src/app/utils/appLink/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { generatePath, Link, LinkProps } from "react-router-dom";
+
+interface AppLinkProps extends Omit<LinkProps, "to"> {
+  to: AppRoute;
+  lang: string;
+  linkId: string | undefined;
+}
+
+export const AppLink: React.FC<React.RefAttributes<HTMLAnchorElement> & AppLinkProps> = props => {
+  const { to, lang, children, linkId, ...rest } = props;
```
FabianHoffmannIntive, line 11: You can make props destructuring one line above:

export const AppLink: React.FC<React.RefAttributes<HTMLAnchorElement> & AppLinkProps> = ({ to, lang, children, linkId, ...rest }) => {

FabianHoffmannIntive (commented):
Please fix all comments

apps/site/src/app/utils/appLink/index.tsx
Reden4005: I wanted to guarantee that my custom AppLink component would have all the <Link> properties available. To conform to the eslint rule, I would have to pinpoint exactly what props I am passing on, rather than spreading. I can add a comment in code why I am turning this rule off.

apps/site/src/app/utils/appLink/index.tsx
Reden4005: Changed.

apps/site-e2e/src/integration/routing/routing.spec.ts
```diff
@@ -0,0 +1,41 @@
+import { AppRoute } from "@patronage-web/shared";
+import { v4 as uuidv4 } from "uuid";
+
+describe("Base routing", () => {
+  const Id = uuidv4();
```
kacperChlon, line 5: Using it this way can lead to have same id everywhere (which couples the tests), hence I would suggest to use it separately in each test instead of having shared id here. 

For now it's not a problem but generally the dynamically generated property like this shouldn't be shared in case of tests :)

Reden4005: Done

apps/site/src/app/utils/routing/index.tsx
```diff
@@ -0,0 +1,27 @@
+import { AppRoute } from "@patronage-web/shared";
+import { Navigate, Outlet, useRoutes } from "react-router-dom";
+
+export const Routing: React.FC = () => {
+  const element = useRoutes([
+    { path: AppRoute.HomePage, element: <h1>Home page</h1> },
```
kacperChlon, line 6: As you can see, here we have many same objects provided for routing, what I suggest is to create common func for them, unify approach and make it look like:

```
export const SampleRouting: React.FC = () => {
  return useRoutes([
    createRoute(AppRoute.Page1, <h1>fallback1</h1>, <h1>page1</h1>, [
      createChildrenRoute(AppRoute.Child1, <h1>child1</h1>),
      createChildrenRoute(AppRoute.Child12, <h1>child12</h1>),
      createChildrenRoute(AppRoute.Child13, <h1>child13</h1>),
      createChildrenRoute(AppRoute.Child14, <h1>child14</h1>)
    ]),
    
    createRoute(AppRoute.Page2, <h1>fallback2</h1>, <h1>page2</h1>),
    createRoute(AppRoute.Page3, <h1>fallback3</h1>, <h1>page3</h1>, [createChildrenRoute(AppRoute.Child3, <h1>Child3</h1>)]),
    createRoute(AppRoute.Page4, <h1>fallback4</h1>, <h1>page4</h1>)
  ])
}
```

Provide proper typing for those functions, fallback should be a component that is used in <Suspense /> component.

apps/site/src/app/utils/appLink/index.tsx
```diff
@@ -0,0 +1,25 @@
+import { AppRoute } from "@patronage-web/shared";
+import { generatePath, Link, LinkProps } from "react-router-dom";
+
+interface AppLinkProps extends Omit<LinkProps, "to"> {
+  to: AppRoute;
+  lang: string;
+  linkId: string | undefined;
+}
+
+export const AppLink: React.FC<React.RefAttributes<HTMLAnchorElement> & AppLinkProps> = props => {
```
kacperChlon, line 10: Was it a scope for your story, to create such component?

apps/site/src/main.tsx
```diff
@@ -1,12 +1,23 @@
...
+
+import { AppLink, Routing } from "./app/utils";
 
 ReactDOM.render(
   <StrictMode>
-    <Box>Patronage App</Box>
+    <BrowserRouter>
+      <Routing />
+      {/* component for testing links, will be removed */}
+      <AppLink linkId='123' to={AppRoute.EditPresentation} lang={i18next.language}>
```
kacperChlon, line 16: We don't want to leave things like this in the code. I know it's temporary but in case of many changes from many people we can omit some of those which we don't want to.

For future reference: 
I suggest to make a usage of jsdocs and provide nice usage explanation with prop description above component https://jsdoc.app/tags-example.html

apps/site/src/app/utils/appLink/index.tsx
Reden4005: I will delete it.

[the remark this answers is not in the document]
Reden4005: Will do.

apps/site/src/main.tsx
Reden4005: I will delete it.

apps/site/src/app/utils/appLink/index.tsx
Reden4005: Component will be deleted.

apps/site/src/app/utils/routing/index.tsx
Reden4005: Code refactored.

apps/site/src/app/utils/routing/index.tsx
```diff
@@ -1,27 +1,39 @@
+import { ReactJSXElement } from "@emotion/react/types/jsx-namespace";
 import { AppRoute } from "@patronage-web/shared";
-import { Navigate, Outlet, useRoutes } from "react-router-dom";
+import { Suspense } from "react";
+import { Outlet, RouteObject, useRoutes } from "react-router-dom";
+
+const createChildrenRoute = (route: AppRoute, component: ReactJSXElement): RouteObject => {
+  return { path: route, element: component };
```
kacperChlon, line 6: Missing fallback for children, it should be same as for the parent

apps/site/src/app/utils/routing/index.tsx
```diff
@@ -1,27 +1,39 @@
...
+};
+
+const createRoute = (
+  route: AppRoute,
+  component: ReactJSXElement,
+  fallback: ReactJSXElement,
```
kacperChlon: fallback type should come from SuspenseProps

apps/site/src/app/utils/routing/index.tsx
```diff
@@ -1,27 +1,39 @@
...
+  children?: RouteObject[]
+): RouteObject => {
+  return {
+    path: route,
+    element: (
+      <Suspense fallback={fallback}>
+        {component}
+        <Outlet />
```
kacperChlon, line 15: Do we need outlet here?

apps/site/src/main.tsx
```diff
@@ -1,22 +1,15 @@
 import "./i18n";
 
-import { AppRoute } from "@patronage-web/shared";
-import i18next from "i18next";
 import { StrictMode } from "react";
 import * as ReactDOM from "react-dom";
 import { BrowserRouter } from "react-router-dom";
 
-import { AppLink, Routing } from "./app/utils";
+import { Routing } from "./app/utils";
```
kacperChlon, line 7: app should have index file from which we should import items => import { Routing } from "./app". Please make an index file with all necessary exports and correct this

libs/shared/src/types/app-routes.ts
```diff
@@ -5,6 +5,5 @@ export enum AppRoute {
   AddPresentation = "add",
   EditPresentation = "edit/:id",
   PresentationForExternalUser = ":id",
-  NotFound = "/notFound",
-  Fallback = "*"
+  PageNotFound = "*"
```
kacperChlon, line 8: We know it's a page because it's in AppRoute, so it's not a big deal but we can read it like 'route not found' which is better than 'route page not found' + documentation of react router dom suggests to name it like NotFound but like I said it's not a big deal, we can leave it :)

apps/site-e2e/src/integration/routing/routing.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+    cy.url().should("include", "/");
+  });
+
+  it(`should display <${AppRoute.Presentation}> page`, () => {
+    cy.visit("localhost:4000/presentation");
+    cy.url().should("include", "/presentation");
+  });
+
+  it(`should display <${AppRoute.Dashboard}> page`, () => {
+    cy.visit("localhost:4000/dashboard");
```
kacperChlon, line 16: Do we need localhost:4000 in those visits?

apps/site-e2e/src/integration/routing/routing.spec.ts
```diff
@@ -0,0 +1,41 @@
+import { AppRoute } from "@patronage-web/shared";
+import { v4 as uuidv4 } from "uuid";
+
+describe("Base routing", () => {
+  it(`should display <${AppRoute.HomePage}> page`, () => {
+    cy.visit("/");
```
kacperChlon, line 6: let's try chaining ;) 

cy.visit("/").url().should("include", "/")

Is it working the same?

Reden4005: Yes, I will change it.

apps/site-e2e/src/integration/routing/routing.spec.ts
Reden4005: I removed localhost:4000. Works fine.

libs/shared/src/types/app-routes.ts
Reden4005: Renamed.

apps/site/src/main.tsx
Reden4005: Done.

apps/site/src/app/utils/routing/index.tsx
Reden4005: Changed to `NonNullable<React.ReactNode> | null`

apps/site/src/app/utils/routing/index.tsx
Reden4005: Children have the same fallback like a parent, because they are in `Outlet` component.

``` tsx
<Suspense fallback={fallback}>
   {component}
   <Outlet />
</Suspense>
```
I checked it.

apps/site/src/app/utils/routing/index.tsx
Reden4005: Yes, if not a child component will not be rendered.

apps/site/src/app/utils/routing/index.tsx
Reden4005: Maybe I should change it to  `{children && <Outlet />}` ?

apps/site-e2e/src/integration/routing/routing.spec.ts
```diff
@@ -0,0 +1,34 @@
+import { AppRoute } from "@patronage-web/shared";
+import { v4 as uuidv4 } from "uuid";
+
+describe("Base routing", () => {
+  it(`should display <${AppRoute.Home}> page`, () => {
+    cy.visit("/").url().should("include", "/");
+  });
+
+  it(`should display <${AppRoute.Presentation}> page`, () => {
+    cy.visit("/presentation").url().should("include", "/presentation");
```
kacperChlon, line 10: please use routes enum everywhere where possible ;)

apps/site/src/app/utils/routing/index.tsx
Reden4005: Maybe I should change it to {children && <Outlet />} ?

Reden4005: Maybe I should change it to `{children && <Outlet />}` ?

apps/site/src/app/utils/routing/index.tsx
kacperChlon: As I mentioned the type here should come from SuspenseProps not ReactNode

apps/site/src/app/utils/routing/index.tsx
kacperChlon: Please consider this two scenarios regarding fallback:
What if our parent will render but our children don't?
What if we'd like to have 10 children and generate them one by one, or all at once?

apps/site/src/app/utils/routing/index.tsx
kacperChlon: Please consider this scenario:
What if we'd like to generate children in specific place, wrapped by something else, different for each parent?

apps/site/src/app/utils/routing/index.tsx
kacperChlon: Try 

{
...
fallback: SuspenseProps["fallback"],
...
}

apps/site/src/app/utils/routing/index.tsx
Reden4005: I moved Outlet component, it is now a part of element in parent route 
` <>
        <h1>Presentation page</h1>
        <Outlet />
      </>,`

apps/site/src/app/utils/routing/index.tsx
Reden4005: Changed.

apps/site/src/app/utils/routing/index.tsx
Reden4005: I moved Outlet component, it is now a part of element in parent route
`<> <h1>Presentation page</h1> <Outlet /> </>`

apps/site/src/app/utils/routing/index.tsx
Reden4005: I wrapped children with Suspense component.

apps/site-e2e/src/integration/routing/routing.spec.ts
Reden4005: Done.

PR #14, opened 2022-01-29
JI:759413 P2022-451 Base navigation header

pawelwojdas:


libs/features/feedback/src/components/nav-header/style.ts
```diff
@@ -0,0 +1,42 @@
...
+    padding: 15px 30px;
+  }
+`;
+
+export const SearchBarPlaceholder = styled.div`
+  order: 1;
+  margin-top: 5px;
+  width: 100%;
+
+  @media (min-width: 900px) {
```
Reden4005, line 27: I just wonder if you can use here https://mui.com/customization/breakpoints/ instead of using hardcoded values?

libs/features/feedback/src/components/nav-header/style.ts
```diff
@@ -0,0 +1,42 @@
...
+  position: fixed;
+  border-bottom: 1px solid lightgrey;
+  width: 100vw;
+`;
+
+export const StyledNavBar = styled.nav`
+  display: flex;
+  align-items: center;
+  flex-wrap: wrap;
+  margin: 0 auto;
```
FabianHoffmannIntive, line 13: justify-content: center ;)

FabianHoffmannIntive (commented):
Please fix all comments

libs/features/feedback/src/components/nav-header/style.ts
kacperChlon: Good point ;) 

But this will come after providing theme config and should be used like this:

```
  [theme.breakpoints.up(breakpoint)]: {
    property: value,
  },
```

hence for now media-query is fine but please make a comment above this line and in every place where media query is used:

// TODO when theme config will be applied - replace with theme breakpoints

apps/site/src/style.ts
```diff
@@ -0,0 +1,7 @@
+import { createGlobalStyle } from "styled-components";
```
kacperChlon, line 1: Please move it to site/utils/create-global-styles.ts

apps/site/src/main.tsx
```diff
@@ -1,12 +1,19 @@
 import "./i18n";
 
-import { Box } from "@mui/material";
+import { NavHeader } from "@patronage-web/features-feedback";
```
kacperChlon, line 3: it's not only feedback feature, for now it contains only feedback elements, however there is a possibility that in the future we would like to dynamically change elements based on selected option/module, hence I would recommend to create generic component in shared directory then use this in site app + import and pass proper configuration like:

```
// this is in features feedback directory (find nice place for it :))
export const navbarConfig = {
    center: [
      <item1 />,
      <item2 />,
      <item3 />
    ],
    end: [
      <item4 />
    ],
  }

// site/src/main.tsx
  import {navbarConfig as feedbackNavbarConfig} from "@features-feedback"
// just sample
  import {navbarConfig as secondNavbarConfig} from "@features-second"

  const configs = {
      [FeatureName.Feedback]:  feedbackNavbarConfig,
      [FeatureName.Feature2]:  secondNavbarConfig,
  }

// in the future it may/will be dynamic
const currentFeature = FeatureName.Feedback

<Navigation
  config={configs[currentFeature]}
/>
```

The navigation will take dynamic configuration and will handle items and styling inside itself.

This is not a full sample, because Navbar will be a part of Layout, however this is not the part of that story.

apps/site/src/main.tsx
```diff
@@ -1,12 +1,19 @@
...
 import { StrictMode } from "react";
 import * as ReactDOM from "react-dom";
 
+import { GlobalStyle } from "./style";
+
 ReactDOM.render(
   <StrictMode>
-    <Box>Patronage App</Box>
+    <GlobalStyle />
+    <NavHeader
```
kacperChlon, line 12: I think Navbar is fine :) but we can stay with NavHeader, is up to you

libs/features/feedback/src/components/nav-header/index.tsx
```diff
@@ -0,0 +1,21 @@
+import { ReactElement } from "react";
+
+import { MenuPlaceholder, SearchBarPlaceholder, StyledNavBar, StyledNavHeader } from "./style";
```
kacperChlon, line 3: please rename it to 'styled' and correct this in readme (I think there is an information to keep styled components in style file

libs/features/feedback/src/components/nav-header/style.ts
```diff
@@ -0,0 +1,42 @@
+import styled from "styled-components";
+
+export const StyledNavHeader = styled.header`
```
kacperChlon, line 3: Please export those as (type)NavHeader, and import them like: 
import * as Styled from './styled'.

Sample:

`<Styled.BorderedNavHeader>...</Styled.BorderedNavHeader>`

also please add the info regarding: naming and importing styled components, into the readme file

libs/features/feedback/src/components/nav-header/style.ts
```diff
@@ -0,0 +1,42 @@
...
+    margin-right: 20px;
+    margin-top: unset;
+    width: auto;
+  }
+`;
+
+export const MenuPlaceholder = styled.div`
+  margin-left: auto;
+
+  @media (min-width: 900px) {
```
kacperChlon, line 39: please add a comment here: 

TODO replace it with theme breakpoints when theme config will be ready

libs/features/feedback/src/components/nav-header/style.ts
pawelwojdas: My bad, thanks 😄

apps/site/src/main.tsx
pawelwojdas: It's a great idea. If component could be more flexible - why not.

apps/site/src/main.tsx
pawelwojdas: I'll change to Navbar, it sounds better

libs/shared/src/components/navbar/style.ts
```diff
@@ -0,0 +1,43 @@
+import { AppBar, Toolbar } from "@mui/material";
+import styled from "styled-components";
```
Reden4005, line 2: Can we use https://mui.com/system/styled/ ?

kacperChlon: Yes, we should for the reasons described in mui docs ;)

pawelwojdas: done

libs/shared/src/components/navbar/index.tsx
```diff
@@ -0,0 +1,18 @@
+import { v4 as uuidv4 } from "uuid";
+
+import { NavbarProps } from "../../types";
```
kacperChlon, line 3: If NavbarProps are not used anywhere else than this file we can leave NavbarProps here

pawelwojdas: It's used for Navbar story as well.

libs/shared/src/components/navbar/index.tsx
```diff
@@ -0,0 +1,18 @@
...
+import * as Styled from "./style";
+
+export const Navbar: React.FC<NavbarProps> = ({ config: { start, center, end }, color = "transparent", variant = "dense" }) => (
+  <Styled.AppBar color={color}>
+    <Styled.Toolbar variant={variant}>
+      {[start, center, end].map(position => (
+        <Styled.SectionContainer key={uuidv4()} sx={position?.customStyles}>
```
kacperChlon, line 10: For such purposes better choice is creating an unique id based on array id combined with some constant name, since key don't need to be globally unique but only for given list or wrapper + keys should be stable to **not lead** to unexpected behaviour like re-rendering etc... please adjust it here and in other places

libs/shared/src/components/navbar/index.tsx
```diff
@@ -0,0 +1,18 @@
+import { v4 as uuidv4 } from "uuid";
+
+import { NavbarProps } from "../../types";
+import * as Styled from "./style";
+
+export const Navbar: React.FC<NavbarProps> = ({ config: { start, center, end }, color = "transparent", variant = "dense" }) => (
+  <Styled.AppBar color={color}>
+    <Styled.Toolbar variant={variant}>
+      {[start, center, end].map(position => (
```
kacperChlon, line 9: please create a const named navItems or sth like that

libs/shared/src/types/navbar.ts
```diff
@@ -0,0 +1,17 @@
+import { AppBarTypeMap, ToolbarTypeMap } from "@mui/material";
+import { MUIStyledCommonProps } from "@mui/material/node_modules/@mui/system";
```
kacperChlon, line 2: I think 
`SxProps from @mui/material`
should be sufficient

package.json
```diff
@@ -61,20 +61,21 @@
     "@emotion/react": "^11.7.1",
     "@mui/icons-material": "^5.2.5",
     "@mui/material": "^5.2.6",
+    "@mui/system": "^5.4.0",
```
kacperChlon, line 64: if this is only needed due to this SxProps type please remove it

libs/shared/src/components/navbar/style.ts
```diff
@@ -0,0 +1,33 @@
...
+});
+
+export const Toolbar = styled(MUIToolbar)({
+  display: "flex",
+  alignItems: "center",
+  alignContent: "space-evenly",
+  flexWrap: "wrap",
+  justifyContent: "space-between",
+  width: "100%",
+  maxWidth: "1536px",
```
kacperChlon, line 17: Please add todo, that in every place where we're using px we should try to use theme.spacing

libs/shared/src/components/navbar/index.tsx
pawelwojdas: done

libs/shared/src/types/navbar.ts
pawelwojdas: done

package.json
pawelwojdas: done

libs/shared/src/components/navbar/style.ts
pawelwojdas: done

FabianHoffmannIntive (approved):

libs/shared/src/components/navbar/style.ts
```diff
@@ -0,0 +1,32 @@
+import { AppBar as MUIAppBar, Toolbar as MUIToolbar } from "@mui/material";
+import { styled } from "@mui/material/styles";
+
+export const AppBar = styled(MUIAppBar)({
```
kacperChlon, line 4: Rename it please to different name, we don't want to use default mui names for styled component :)

PR #17, opened 2022-02-01
JI:759414 P2022-452 Navigation breadcrumbs

krysia1:


libs/shared/src/components/logo/index.tsx
```diff
@@ -0,0 +1,5 @@
+import * as Styled from "./styled";
+
+export const Logo: React.FC = () => {
+  return <Styled.BasicLogo />;
+};
```
Adys23, line 5: You can move code from _logo/styled.ts_ to _nav-breadcrumbs/styled.ts_ and return `<Styled.BasicLogo />` directly in _nav-breadcrumbs/index.tsx_ instead of `<Logo />`

libs/features/feedback/src/components/nav-breadcrumbs/styled.ts
```diff
@@ -0,0 +1,15 @@
...
+
+// TODO when theme config will be applied - replace with theme color, weight, size
+export const MediumTypography = styled(Typography)`
+  color: rgb(126, 126, 126);
+  font-weight: 500;
+  font-size: 14px;
+`;
+
+// TODO when theme config will be applied - replace with theme color
+export const ArrowSeparator = styled(NavigateNextIcon)`
```
Adys23, line 13: Types for components are missing

krysia1: It is already inheriting types from `styled` and `NavigateNextIcon`
image
How would you type it?

libs/features/feedback/src/components/nav-breadcrumbs/index.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+import * as Styled from "./styled";
+
+export interface NavBreadcrumbsProps {
+  presentationName?: string;
+}
+
+export const NavBreadcrumbs: React.FC<NavBreadcrumbsProps> = ({ presentationName }) => {
+  return (
+    <Breadcrumbs aria-label='breadcrumb' separator={<Styled.ArrowSeparator />}>
```
kacperChlon, line 12: If we want to use aria-labeling please add it to translations

libs/shared/src/components/logo/index.tsx
kacperChlon: Yes, there is no need to wrap just styled component, this is also one of the reasons why styled components are useful :)

libs/shared/src/components/logo/styled.ts
```diff
@@ -0,0 +1,12 @@
+import { ImageOutlined as ImageOutlinedIcon } from "@mui/icons-material";
+import styled from "styled-components";
+
+// TODO when theme config will be applied - replace with theme color
+export const BasicLogo = styled(ImageOutlinedIcon).attrs({
+  fontSize: "large"
+})`
```
kacperChlon, line 7: To unify approach in styling instead of _template literal_, could you use css _css-in-js_?

libs/shared/src/components/logo/index.tsx
krysia1: I see. I wasn't sure where to put it, thanks :D

libs/shared/src/components/logo/styled.ts
krysia1: You mean I should change this one to _template literal_ ? Because I think that's what we were using

libs/features/feedback/src/components/nav-breadcrumbs/styled.ts
Adys23: sorry, I haven't spot that. I am using `React.FC` type and I get used to that.

libs/features/feedback/src/components/nav-breadcrumbs/styled.ts
```diff
@@ -0,0 +1,15 @@
+import { NavigateNext as NavigateNextIcon } from "@mui/icons-material";
+import { Typography } from "@mui/material";
+import styled from "styled-components";
```
Reden4005, line 3: Can we use https://mui.com/system/styled/ ?

libs/shared/src/components/logo/styled.ts
kacperChlon: No no :) I mean **instead of** template-literal we should use [css-in-js](https://cssinjs.org/?v=v10.9.0) approach everywhere.

libs/features/feedback/src/components/nav-breadcrumbs/styled.ts
kacperChlon: https://github.com/intive/patronage22-web-wroclaw/pull/14#discussion_r797532267

libs/shared/src/components/logo/styled.ts
krysia1: oh ok

libs/features/feedback/src/components/nav-breadcrumbs/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+export interface NavBreadcrumbsProps {
+  presentationName?: string;
+}
+
+export const NavBreadcrumbs: React.FC<NavBreadcrumbsProps> = ({ presentationName }) => {
+  const { t } = useTranslation();
+
+  return (
+    <Breadcrumbs aria-label={t("breadcrumb")} separator={<Styled.ArrowSeparator />}>
+      <Styled.AlignedCenterLink href='/'>
```
kacperChlon, line 15: please use Routing [enum ](https://github.com/intive/patronage22-web-wroclaw/blob/develop/libs/shared/src/types/app-routes.ts) for such paths

PR #23, opened 2022-02-05
JI:761104 P2022-685 Create Localized Link component

Reden4005:
`LocalizedLink` component is a wrapper for `Link` component which generate a path with querystring with language.


apps/site/src/app/utils/localized-link.tsx
```diff
@@ -0,0 +1,27 @@
...
+  to,
+  children,
+  params,
+  ...rest
+}) => (
+  <Link
+    // Disabled, to pass all props from Link component
+    // eslint-disable-next-line react/jsx-props-no-spreading
+    {...rest}
+    to={`${generatePath(`${to instanceof Array ? to.join("/") : to}`, params)}?lang=${i18next.language}`}
```
kacperChlon, line 23: Instead of using i18next here please use: https://react.i18next.com/latest/usetranslation-hook

apps/site/src/app/utils/localized-link.tsx
```diff
@@ -0,0 +1,27 @@
+import "../../i18n";
```
kacperChlon, line 1: LocalizedLink is a component, not util and it will be used across all modules in whole application

apps/site/src/app/utils/localized-link.tsx
```diff
@@ -0,0 +1,27 @@
...
+import i18next from "i18next";
+import { generatePath, Link, LinkProps, Params } from "react-router-dom";
+
+interface LocalizedLinkProps extends Omit<LinkProps, "to"> {
+  to: AppRoute | AppRoute[];
+  linkId?: string;
+  params?: Params;
+}
+
+export const LocalizedLink: React.FC<React.RefAttributes<HTMLAnchorElement> & LocalizedLinkProps> = ({
```
kacperChlon, line 13: Please import RefAttributes form 'react' library

apps/site/src/app/utils/localized-link.tsx
```diff
@@ -0,0 +1,27 @@
...
+  to,
+  children,
+  params,
+  ...rest
```
kacperChlon, line 17: We should specify only those properties that we want to use and are important for proper functioning.

apps/site/src/app/utils/localized-link.tsx
kacperChlon, line 23: please move this to separate function within this component

apps/site/src/app/utils/localized-link.tsx
```diff
@@ -0,0 +1,27 @@
+import "../../i18n";
+
+import { AppRoute } from "@patronage-web/shared";
+import i18next from "i18next";
+import { generatePath, Link, LinkProps, Params } from "react-router-dom";
+
+interface LocalizedLinkProps extends Omit<LinkProps, "to"> {
+  to: AppRoute | AppRoute[];
+  linkId?: string;
+  params?: Params;
```
kacperChlon, line 10: please rename it to routerParams

apps/site/src/app/utils/localized-link.tsx
```diff
@@ -0,0 +1,27 @@
+import "../../i18n";
+
+import { AppRoute } from "@patronage-web/shared";
+import i18next from "i18next";
+import { generatePath, Link, LinkProps, Params } from "react-router-dom";
+
+interface LocalizedLinkProps extends Omit<LinkProps, "to"> {
+  to: AppRoute | AppRoute[];
+  linkId?: string;
```
kacperChlon, line 9: why do we need it here?

apps/site/src/app/utils/localized-link.tsx
Reden4005: Done.

apps/site/src/app/utils/localized-link.tsx
Reden4005: Removed to libs/shared/component.

apps/site/src/app/utils/localized-link.tsx
Reden4005: According to the next comment I don't need it anymore.

apps/site/src/app/utils/localized-link.tsx
Reden4005: Removed.

apps/site/src/app/utils/localized-link.tsx
Reden4005: Done.

apps/site/src/app/utils/localized-link.tsx
Reden4005: Done.

apps/site/src/app/utils/localized-link.tsx
Reden4005: Removed.

libs/shared/src/components/localized-link.tsx
```diff
@@ -0,0 +1,16 @@
+import { AppRoute } from "@patronage-web/shared";
```
kacperChlon, line 1: For components we are following given approach:
.../component-name/index.tsx

kacperChlon (approved):

libs/shared/src/components/localized-link.tsx
Reden4005: Done.

PR #24, opened 2022-02-07
JI:762546 P2022-954 Replace MUILink with LocalizedLink

krysia1:


libs/features/feedback/src/components/nav-breadcrumbs/styled.ts
```diff
@@ -1,5 +1,5 @@
 import { ImageOutlined as MUIImageOutlinedIcon, NavigateNext as MUINavigateNextIcon } from "@mui/icons-material";
```
kacperChlon, line 1: It's not a big deal but just for your information ;) if the names are not collide with each other we don't have to import them as CustomNameImport :) Don't change it, just remember ;)

libs/features/feedback/src/components/nav-breadcrumbs/styled.ts
```diff
@@ -20,7 +20,7 @@ export const BasicLogo = styled(MUIImageOutlinedIcon)({
   fontSize: "1.875rem"
 });
 
-export const AlignedCenterLink = styled(MUILink)({
+export const AlignedCenterWrapper = styled("div")({
```
kacperChlon, line 23: Can we change div to Box? We want to follow MaterialUI as much as possible

libs/features/feedback/src/components/nav-breadcrumbs/styled.ts
krysia1: ok, thanks :)

libs/features/feedback/src/components/nav-breadcrumbs/index.tsx
```diff
@@ -13,9 +13,11 @@ export const NavBreadcrumbs: React.FC<NavBreadcrumbsProps> = ({ presentationName
 
   return (
     <Breadcrumbs aria-label={t("breadcrumb")} separator={<Styled.ArrowSeparator />}>
-      <Styled.AlignedCenterLink href={AppRoute.Home}>
-        <Styled.BasicLogo />
-      </Styled.AlignedCenterLink>
+      <LocalizedLink to={AppRoute.Home}>
```
Reden4005, line 16: Localized Link - we don't use it like this anymore. You should import `PagePath from "@patronage-web/shared"`,  and use it like this: 
`<LocalizedLink to={PagePath.Home}>`

krysia1: I'm literally changing it right now ;P thanks :D

kacperChlon (approved):

PR #25, opened 2022-02-08
Feature/p2022 972 refactor localized link component

Reden4005:
Refactor localized link component.


libs/shared/src/types/app-routes.ts
```diff
@@ -7,3 +7,21 @@ export enum AppRoute {
...
+  Home = "home",
+  Presentation = "presentation",
+  AddPresentation = "add-presentation",
+  EditPresentation = "edit-presentation",
+  PresentationForExternalUser = "external-user-presentation",
+  Dashboard = "dashboard"
+}
+
+export const PAGE_PATHS: Record<PagePath, AppRoute | AppRoute[]> = {
+  [PagePath.Dashboard]: AppRoute.Dashboard,
```
pawelwojdas, line 21: PAGE_PATHS are constants, not types. Shouldn't we move it to other directory?

libs/shared/src/components/localized-link/index.tsx
```diff
@@ -1,16 +1,19 @@
-import { AppRoute } from "@patronage-web/shared";
 import { useTranslation } from "react-i18next";
-import { generatePath, Link, Params } from "react-router-dom";
+import { Link, Params } from "react-router-dom";
+
+import { PAGE_PATHS, PagePath } from "../../types";
+import { createPath } from "../../utils";
```
pawelwojdas, line 5: Shouldn't we import as: `import { createPath, PAGE_PATHS, PagePath } from "@patronage-web/shared";`

libs/shared/src/utils/create-path.ts
```diff
@@ -0,0 +1,9 @@
+import { generatePath, Params } from "react-router-dom";
+
+import { AppRoute, SupportedLanguage } from "../types";
+
+export const createPath = (path: AppRoute | AppRoute[], params?: Params, language?: string) => {
+  const lang = language?.length && language !== SupportedLanguage.En ? `?lang=${language}` : "";
+  const basePath = path instanceof Array ? path.join("/") : path;
```
kacperChlon, line 7: Please add an empty space before return

libs/shared/src/types/app-routes.ts
kacperChlon: Yes, we can ;) to constants/index.ts

libs/shared/src/components/localized-link/index.tsx
kacperChlon: No, and this is also set as a rule in eslint:
image

It ensures us we won't import parent index into child within same location.

libs/shared/src/types/app-routes.ts
Reden4005: Done.

libs/shared/src/components/localized-link/index.tsx
Reden4005: Indeed.

libs/shared/src/utils/create-path.ts
Reden4005: Done.

PR #27, opened 2022-02-08
JI:762802 P2022-980 Implement footer according to AC

malmon305:


kacperChlon:
Please rename PR using tampermonkey script

libs/shared/src/components/footer/style.ts
```diff
@@ -0,0 +1,35 @@
+import { AppBar as MUIAppBar, Toolbar as MUIToolbar } from "@mui/material";
+import { styled } from "@mui/material/styles";
+
+export const AppBar = styled(MUIAppBar)({
```
kacperChlon, line 4: Please change all the names that looks the same as mui default, we don't want to use default mui names for styled component because someone else can do that as well and we can have name collision.

Could you do the same thing for navbar?
https://github.com/intive/patronage22-web-wroclaw/blob/develop/libs/shared/src/components/navbar/style.ts

libs/shared/src/components/footer/style.ts
```diff
@@ -0,0 +1,35 @@
...
+  flexDirection: "column",
+  bottom: "0",
```
kacperChlon, line 6: can be a number

apps/site/src/main.tsx
```diff
@@ -11,6 +13,7 @@ ReactDOM.render(
     <BrowserRouter>
       <Routing />
     </BrowserRouter>
+    <Footer config={feedbackFooterConfig} />
```
kacperChlon, line 16: Please remove it from main component, it will be moved to Layout component (in upcoming sprint) with navbar.

libs/shared/src/types/footer.ts
```diff
@@ -0,0 +1,17 @@
...
+  elements: JSX.Element[];
+}
+
+export enum FooterSectionPosition {
+  Start = "start",
+  End = "end"
+}
+
+export interface FooterProps {
+  color?: AppBarTypeMap["props"]["color"];
```
kacperChlon, line 14: Make a usage of typescript **Pick<>** for color and variant

libs/shared/src/types/footer.ts
```diff
@@ -0,0 +1,17 @@
+import { AppBarTypeMap, SxProps, ToolbarTypeMap } from "@mui/material";
```
kacperChlon, line 1: Can we move those styles to footer component directory? Those types are not global shared types but are relevant only for Footer component.  It will be easier to track changes and keep all in one place if such types will be in same direactory as shared component.

Could you do that for navbar as well?

libs/shared/src/components/footer/style.ts
```diff
@@ -0,0 +1,35 @@
...
+  maxWidth: theme.spacing(192),
+  margin: theme.spacing(1, 0),
+  marginTop: "auto"
+}));
+
+export const SectionContainer = styled("div")({
+  display: "flex"
+});
+
+export const SectionItem = styled("div")(({ theme }) => ({
```
kacperChlon, line 28: https://github.com/intive/patronage22-web-wroclaw/pull/24#discussion_r801499113

libs/shared/src/components/footer/style.ts
```diff
@@ -0,0 +1,35 @@
...
+  alignContent: "space-evenly",
+  flexWrap: "wrap",
+  justifyContent: "space-between",
+  width: "100%",
+  maxWidth: theme.spacing(192),
+  margin: theme.spacing(1, 0),
+  marginTop: "auto"
+}));
+
+export const SectionContainer = styled("div")({
```
kacperChlon, line 24: https://github.com/intive/patronage22-web-wroclaw/pull/27/files#r802165978

libs/shared/src/types/footer.ts
malmon305: I'm not sure I understand how you would like me to use `Pick<>`? Could you explain? 
From what I understood `Pick<>` allows creation of new types from existing ones, by selecting some keys from existing type. How is that helping me here? If I will take `Pick<AppBarTypeMap["props"], "color">` it's an interface with property color not union that I need here `"inherit" | "primary" | "secondary" | "default" | "transparent" | undefined`.

libs/shared/src/components/footer/style.ts
malmon305: done

libs/shared/src/components/footer/style.ts
malmon305: done

apps/site/src/main.tsx
malmon305: done

libs/shared/src/types/footer.ts
malmon305: done

libs/shared/src/components/footer/style.ts
malmon305: done

libs/shared/src/types/footer.ts
kacperChlon: In case of interfaces to inherit some types/properties we should extend the interface.

Try 
interface FooterProps extends Pick<AppBarProps, "color"> {
  ...,
}

libs/shared/src/components/footer/style.ts
```diff
@@ -0,0 +1,50 @@
...
+  elements: JSX.Element[];
+}
+
+export enum FooterSectionPosition {
+  Start = "start",
+  End = "end"
+}
+
+export interface FooterProps {
+  color?: AppBarTypeMap["props"]["color"];
```
kacperChlon, line 15: please use Pick

libs/shared/src/types/footer.ts
malmon305: Thanks for the tip. Fixed

libs/shared/src/components/navbar/style.ts
```diff
@@ -1,13 +1,29 @@
...
+export enum NavbarSectionPosition {
+  Start = "start",
+  Center = "center",
+  End = "end"
+}
+
+export interface NavbarProps {
+  color?: AppBarTypeMap["props"]["color"];
+  config: Partial<Record<NavbarSectionPosition, NavbarSection>>;
+  variant?: ToolbarTypeMap["props"]["variant"];
```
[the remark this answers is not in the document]
malmon305: Done

kacperChlon (approved):

PR #28, opened 2022-02-09
JI:759433 P2022-465 Share or save presentation

pawelwojdas:


libs/features/feedback/src/components/save-button/index.tsx
```diff
@@ -0,0 +1,15 @@
...
+export const SaveButton = () => {
+  const { t } = useTranslation();
+  const navigate = useNavigate();
+
+  const onClickHandler = () => {
+    // TODO add new presentation
+    navigate(createPath(AppRoute.Dashboard));
+  };
+
+  return <BasicButton text={t("save")} onClick={onClickHandler} />;
```
Adys23, line 14: Instead of specific component for save-button, can't we use BasicButton directly in the place we want to use it and simply pass text and handleClick as props?

libs/features/feedback/src/components/share-button/index.tsx
```diff
@@ -0,0 +1,11 @@
+import ShareOutlinedIcon from "@mui/icons-material/ShareOutlined";
+import { IconButton } from "@patronage-web/shared";
+
+// TODO apply openDialogHandler when redux will be applied
+export const ShareButton: React.FC = () => {
+  return (
+    <IconButton>
+      <ShareOutlinedIcon />
+    </IconButton>
+  );
```
Adys23, line 10: Similar might be done here: IconButton which accepts type of Icon you would like to use as a prop?

libs/features/feedback/src/components/save-button/index.tsx
pawelwojdas: Yes, when component for those buttons will be ready, I'll move them.

libs/shared/src/components/styled/basic-icon-button/index.tsx
```diff
@@ -0,0 +1,14 @@
+import { IconButtonProps, useMediaQuery, useTheme } from "@mui/material";
```
kacperChlon, line 1: We should use styled folder in components directory only if the whole component can be just styled component without any additional logic.

libs/shared/src/components/styled/basic-button/index.tsx
```diff
@@ -0,0 +1,24 @@
+import { ButtonProps, useMediaQuery, useTheme } from "@mui/material";
+
+import * as Styled from "./style";
+
+export interface BasicButtonProps extends ButtonProps {
+  text: string;
+}
+
+export const BasicButton: React.FC<BasicButtonProps> = ({ disabled, text, onClick, variant = "outlined" }) => {
```
kacperChlon, line 9: same comments as in case of basic icon button

libs/shared/src/hooks/use-screen.ts
```diff
@@ -0,0 +1,13 @@
+import { useMediaQuery, useTheme } from "@mui/material";
+
+export const useScreen = () => {
```
kacperChlon, line 3: What do you think about renaming it to useScreenSize

libs/shared/src/utils/get-theme-config.tsx
```diff
@@ -3,16 +3,24 @@ import { createTheme } from "@mui/material/styles";
 
 import { ThemeMode } from "../types";
 
+export const generalTheme = createTheme({
```
kacperChlon, line 6: What do you think about **commonStyles** instead of generalTheme? ;)

libs/features/feedback/src/components/save-button/index.tsx
```diff
@@ -0,0 +1,20 @@
+import { AppRoute, BasicButton, createPath, useScreen } from "@patronage-web/shared";
```
kacperChlon, line 1: I think there is no need for such specific button, instead we can create more generic button like BaseButton and give it a props: children, onClick and variant. then we will be able to reuse it anywhere.

libs/features/feedback/src/components/save-button/index.tsx
```diff
@@ -0,0 +1,20 @@
+import { AppRoute, BasicButton, createPath, useScreen } from "@patronage-web/shared";
+import { useTranslation } from "react-i18next";
+import { useNavigate } from "react-router-dom";
+
+export const SaveButton = () => {
```
kacperChlon, line 5: lack of component typing

libs/shared/src/hooks/use-screen.ts
pawelwojdas: It's really good idea, it'll be easier to realize what's hook for

libs/shared/src/hooks/use-screen.ts
```diff
@@ -0,0 +1,13 @@
+import { useMediaQuery, useTheme } from "@mui/material";
+
+export const useScreen = () => {
+  const theme = useTheme();
+
+  const sizes = {
```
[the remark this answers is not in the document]
pawelwojdas: done

FabianHoffmannIntive (dismissed):
Please resolve Kacper's comments

libs/features/feedback/src/components/save-button/index.tsx
pawelwojdas: done

libs/features/feedback/src/components/save-button/index.tsx
pawelwojdas: I removed SaveButton

libs/features/feedback/src/components/share-button/index.tsx
```diff
@@ -0,0 +1,14 @@
+import ShareOutlinedIcon from "@mui/icons-material/ShareOutlined";
+import { IconButton } from "@mui/material";
+import { useScreen } from "@patronage-web/shared";
+
+// TODO apply handleClick when redux and Dialog component will be applied
+export const ShareButton: React.FC = () => {
```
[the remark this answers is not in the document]
pawelwojdas: I removed ShareButton. Component will be implemented when parent component (where this button and SaveButton should be) will be ready

libs/shared/src/components/base-button/index.tsx
```diff
@@ -0,0 +1,24 @@
+import { ButtonProps, IconButton } from "@mui/material";
+
+import { useScreenSize } from "../../hooks";
+import { ButtonType } from "../../types";
```
kacperChlon, line 4: Please move this enum to this component since this is strictly related

libs/shared/src/components/base-button/index.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+export interface BaseButtonProps extends Pick<ButtonProps, "children" | "onClick" | "variant"> {
+  type: ButtonType;
+}
+
+export const BaseButton: React.FC<BaseButtonProps> = ({ children, onClick, type, variant }) => {
+  const { isMobile } = useScreenSize();
+  const buttonSize = isMobile ? "small" : "medium";
+
+  return type === ButtonType.Basic ? (
```
kacperChlon, line 15: please use object instead of ternary operator and return based on type

libs/shared/src/components/styled/basic-icon-button/index.tsx
pawelwojdas: done

libs/shared/src/components/base-button/index.tsx
pawelwojdas: Ok, done

libs/shared/src/components/base-button/index.tsx
pawelwojdas: Good idea, always better performance 😄

kacperChlon (approved):

pawelwojdas:
> Please resolve Kacper's comments

They are resolved already.

PR #32, opened 2022-02-14
JI:763128 P2022-1048 Create initial pages

krysia1:


apps/site/src/app/pages/dashboard/styled.ts
```diff
@@ -0,0 +1,10 @@
+import { Box } from "@mui/material";
+import { styled } from "@mui/material/styles";
+
+// sample styles
+export const CenteredWhiteWrapper = styled(Box)({
```
kacperChlon, line 5: First of all, if you have so many same elements it can be moved to one place and be exported from there plus such styling applied to each page should be done by one component (wrapper component for all pages). And in our case it's gonna be https://tracker.intive.com/jira/browse/P2022-942

apps/site/src/app/pages/not-found/index.tsx
```diff
@@ -0,0 +1,17 @@
+import { CardMedia } from "@mui/material";
+import { NotFoundImage } from "@patronage-web/features-feedback";
```
kacperChlon, line 2: Since it will be generic page for all not found cases the image should be placed in shared directory ;)

apps/site/src/app/pages/not-found/index.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+import * as Styled from "./styled";
+
+export const NotFoundPage: React.FC = () => {
+  return (
+    // sample data
```
kacperChlon, line 8: instead of such comment you can provide sth like:
// TO DO - replace when proper data will be delivered/committed

apps/site/src/app/pages/presentation-for-external-user/index.tsx
```diff
@@ -0,0 +1,10 @@
+import * as Styled from "./styled";
+
+export const PresentationForExternalUserPage: React.FC = () => {
```
kacperChlon, line 3: I would rather use name ExternalUserPresentationPage

apps/site/src/app/utils/routing.tsx
```diff
@@ -1,7 +1,17 @@
-import { AppRoute } from "@patronage-web/shared";
-import React, { Suspense, SuspenseProps } from "react";
+import { AppRoute, Loader } from "@patronage-web/shared";
+import { lazy, Suspense, SuspenseProps } from "react";
 import { Outlet, RouteObject, useRoutes } from "react-router-dom";
 
+const Homepage = lazy(() => import("../pages").then(module => ({ default: module.Homepage })));
```
kacperChlon, line 5: What is the reason for doing `.then({default: ...})`?

apps/site/src/app/utils/routing.tsx
```diff
@@ -18,20 +28,20 @@ const createChildrenRoute = (route: AppRoute, component: JSX.Element) => (fallba
...
       </>,
-      <h1>...loading</h1>,
+      <Loader />,
       [
-        createChildrenRoute(AppRoute.AddPresentation, <h1>Add presentation page</h1>),
-        createChildrenRoute(AppRoute.EditPresentation, <h1>Edit presentation page</h1>),
-        createChildrenRoute(AppRoute.PresentationForExternalUser, <h1>Page for external user</h1>)
+        createChildrenRoute(AppRoute.AddPresentation, <AddPresentationPage />),
+        createChildrenRoute(AppRoute.EditPresentation, <EditPresentationPage />),
+        createChildrenRoute(AppRoute.PresentationForExternalUser, <PresentationForExternalUserPage />)
```
kacperChlon, line 43: Please change this AppRoute to ExternalUserPresentation as well

libs/shared/src/components/loader/index.tsx
```diff
@@ -0,0 +1,11 @@
+import { CircularProgress } from "@mui/material";
+
+import * as Styled from "./styled";
+
+export const Loader: React.FC = () => {
```
kacperChlon, line 5: Please add to this components type property and extend it to based on type return Circular or Linear loader ;)

libs/shared/src/components/loader/styled.ts
```diff
@@ -0,0 +1,10 @@
+import { Box } from "@mui/material";
+import { styled } from "@mui/material/styles";
+
+export const CenteredWrapper = styled(Box)({
```
kacperChlon, line 4: LoaderWrapper maybe? ;)

apps/site/src/app/pages/dashboard/styled.ts
krysia1: I agree, I just wanted to add sample elements to the pages

apps/site/src/app/pages/not-found/index.tsx
krysia1: kk

apps/site/src/app/pages/not-found/index.tsx
krysia1: ok

apps/site/src/app/pages/presentation-for-external-user/index.tsx
krysia1: Yeah, sounds better :)

apps/site/src/app/utils/routing.tsx
krysia1: We cannot do that because we're exporting our components as `export const` instead of `export default` :/

apps/site/src/app/utils/routing.tsx
krysia1: kk

libs/shared/src/components/loader/index.tsx
krysia1: Will do :)

libs/shared/src/components/loader/styled.ts
krysia1: Yeeah 🤔

apps/site/src/app/utils/routing.tsx
kacperChlon: Ok, so for pages we can make an exception and export them as defaults. It's reasonable since they won't be used anywhere else and each page is sort of main wrapper/component for all its content :) Please use default exports and remove index.ts from pages directory

krysia1: ok :)

apps/site/src/app/pages/homepage/index.tsx
```diff
@@ -4,7 +4,7 @@ import { useTranslation } from "react-i18next";
 
 import * as Styled from "./styled";
 
-export const Homepage: React.FC = () => {
+const Homepage: React.FC = () => {
   const { t } = useTranslation(TranslationNamespace.Common);
 
   return (
```
kacperChlon, line 10: please create **home** component in **views** in **features-feedback** and import here **HomeView** as **HomeFeedbackView**

apps/site/src/app/pages/homepage/index.tsx
```diff
@@ -4,7 +4,7 @@ import { useTranslation } from "react-i18next";
 
 import * as Styled from "./styled";
 
-export const Homepage: React.FC = () => {
+const Homepage: React.FC = () => {
   const { t } = useTranslation(TranslationNamespace.Common);
```
kacperChlon, line 8: Is this necessary? ;)

apps/site/src/app/pages/not-found/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+const NotFoundPage: React.FC = () => {
+  return (
+    // TODO - replace when proper data will be delivered
+    <>
+      <Styled.MediumImageWrapper>
+        <CardMedia component='img' alt='404' src={NotFoundImage} />
+      </Styled.MediumImageWrapper>
+
+      <h1>Page Not Found</h1>
```
kacperChlon, line 14: please add this to translations

libs/shared/src/components/loader/index.tsx
```diff
@@ -0,0 +1,21 @@
+import { CircularProgress, LinearProgress } from "@mui/material";
+
+import { LoaderType } from "../../types";
+import * as Styled from "./styled";
+
+export interface LoaderProps {
+  loaderType: LoaderType;
```
kacperChlon, line 7: no need to name it loaderType since you are passing this to Loader component as LoaderProps, hence type is enough :)

libs/shared/src/components/loader/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+}
+
+export const Loader: React.FC<LoaderProps> = ({ loaderType }) => {
+  const PROGRESS_TYPE = {
+    circular: <CircularProgress color='inherit' />,
```
kacperChlon, line 12: please use enum as keys :)

apps/site/src/app/pages/homepage/index.tsx
krysia1: Ok, I've done what I think you've meant ;)

apps/site/src/app/pages/homepage/index.tsx
krysia1: I guess it's not

libs/shared/src/components/loader/index.tsx
```diff
@@ -0,0 +1,21 @@
+import { CircularProgress, LinearProgress } from "@mui/material";
+
+import { LoaderType } from "../../types";
```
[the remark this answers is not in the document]
krysia1: ok

libs/shared/src/components/loader/index.tsx
krysia1: Yeah, you're right

libs/shared/src/components/loader/index.tsx
krysia1: kk

libs/shared/src/types/http-status.ts
```diff
@@ -0,0 +1,3 @@
+export enum HttpStatus {
+  NotFound = 404
```
kacperChlon, line 2: please use string here :) then I think you won't need to wrap it with template literals in not found view

libs/translations/src/common/en.json
```diff
@@ -2,5 +2,8 @@
   "homepage": {
     "newPresentationDescription": "Placeholder text for ActionCard to create new presentation on Homepage",
     "newPresentationButton": "New presentation"
+  },
+  "notFoundPage": {
+    "info": "Page Not Found"
```
kacperChlon, line 7: 'title' maybe? ;)

apps/site/src/app/pages/not-found/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+import * as Styled from "./styled";
+
+const NotFoundPage: React.FC = () => {
+  const { t } = useTranslation();
+
+  return (
+    // TODO - replace when proper data will be delivered
+    <>
+      <Styled.MediumImageWrapper>
+        <CardMedia component='img' alt={`${HttpStatus.NotFound}`} src={NotFoundImage} />
```
kacperChlon, line 14: Do we need to wrap HttpStatus.NotFound in template literal?

libs/shared/src/types/http-status.ts
krysia1: Yes, I thought that maybe we will expand this enum and use it when checking responses' status. That's why it's number

apps/site/src/app/pages/not-found/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { NotFoundImage } from "@patronage-web/shared";
+
+import * as Styled from "./styled";
+
+const NotFoundPage: React.FC = () => {
+  return (
+    // TODO - replace when proper data will be delivered
+    <>
+      <Styled.MediumImageWrapper>
+        <CardMedia component='img' alt='404' src={NotFoundImage} />
```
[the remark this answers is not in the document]
krysia1: done

libs/translations/src/common/en.json
krysia1: ok

kacperChlon (approved):

PR #43, opened 2022-02-22
JI:765256 P2022-1249 1. Setup basic store with sample reducer

krysia1:


libs/shared/data/feedback/mocks/sample-reducer/selectors.ts
```diff
@@ -0,0 +1 @@
+export const selectSampleText = (state: { sample: { text: string } }) => state.sample.text;
```
kacperChlon, line 1: I would avoid destructuring here, sometimes you don't know if you've got the value, moreover what's the point in destructuring if you are using state.sample.text? ;)

libs/shared/data/feedback/mocks/sample-reducer/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from "./actions";
```
kacperChlon, line 1: Why those files are in mocks directory? :)

krysia1: Great question, I guess I had a reason... but know I don't know what was it 🤣

libs/shared/data/feedback/mocks/sample-reducer/selectors.ts
krysia1: ok

libs/shared/data/features/sample-reducer/actions.ts
```diff
@@ -0,0 +1,3 @@
+import { sampleSlice } from "./sample-slice";
```
kacperChlon, line 1: this all files shouldn't be in features directory, this is not 'features, it's features/sample/....

Please move all files related to sample-feature to features/sample dir

tsconfig.base.json
```diff
@@ -23,6 +23,7 @@
     "paths": {
       "@patronage-web/features-feedback": ["libs/features/feedback/src/index.ts"],
       "@patronage-web/shared": ["libs/shared/src/index.ts"],
+      "@patronage-web/shared-data": ["libs/shared/data/index.ts"],
```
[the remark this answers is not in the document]
kacperChlon: No, it means looks good to me ;)

libs/shared/data/features/sample-reducer/selectors.ts
```diff
@@ -0,0 +1,4 @@
+// eslint-disable-next-line import/no-cycle
```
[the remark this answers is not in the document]
krysia1: ok

libs/shared/data/features/sample-reducer/actions.ts
krysia1: I see, ok done 😸

kacperChlon (approved):

PR #56, opened 2022-03-05
JI:760821 P2022-647 Login page (part 1)

Adys23:


libs/shared/src/components/form-field/index.tsx
```diff
@@ -46,10 +53,13 @@ export const FormField: React.FC<FormFieldProps> = ({
   };
 
   return (
-    <Styled.Field>
-      {renderField({ type, name, value, handleChange, variant, errors: fieldErrors, label, rows })}
-      {!hideEditIcon && <Edit />}
-      {renderHelperText(fieldErrors, helperText)}
-    </Styled.Field>
+    <>
+      {showDescription && <Typography>{description}</Typography>}
```
kacperChlon, line 57: maybe description && <Typography>{description}</Typography>? ;)

libs/shared/src/components/form/index.tsx
```diff
@@ -38,6 +45,8 @@ export const Form: React.FC<FormProps> = ({
   showCancelButton,
   submitButtonText,
   cancelButtonText,
+  submitButtonIcon,
```
kacperChlon, line 48: to avoid such properties (in 2 weeks we will have submitButtonColor, submitButtonSomethinngElse...), let's create property: 
```

FormButton {
   condition, text, icon,
}

customButtons?: FormButton[]
```

libs/shared/src/components/form/index.tsx
```diff
@@ -70,21 +79,25 @@ export const Form: React.FC<FormProps> = ({
...
+    { condition: showSubmitButton, text: submitButtonText || t("submit"), action: handleSubmit, icon: submitButtonIcon },
+    { condition: showCancelButton, text: cancelButtonText || t("cancel"), action: handleCancel, icon: cancelButtonIcon }
   ];
 
   return (
     <FormProvider {...methods}>
       <Styled.Form className={className}>
         {title && <Typography variant='h3'>{title}</Typography>}
 
+        {customField}
```
kacperChlon, line 91: let's do sth like:
```
{title ...

{customField}
{renderFields()}
{children}

{formButtons...
```

libs/shared/src/components/form/index.tsx
```diff
@@ -23,10 +26,14 @@ export interface FormProps {
...
   submitButtonText?: string;
   cancelButtonText?: string;
+  submitButtonIcon?: ReactNode;
+  cancelButtonIcon?: ReactNode;
   className?: string;
 }
 
 export const Form: React.FC<FormProps> = ({
+  children,
+  customField,
```
kacperChlon, line 36: Why do we need customField? I don't know if it's some customFormField or what kind of a field `+` Form component should be only responsible for rendering form fields not some additional one

libs/shared/src/components/login/styled.ts
```diff
@@ -0,0 +1,51 @@
+import { Box, Card, styled, Typography } from "@mui/material";
+
+import { Form } from "../form";
+
+export const Container = styled(Box)(({ theme }) => ({
```
kacperChlon, line 5: Please use more descriptive names, I can have 1000 containers in the application as well as 1000 titles and Subtitles, please rename those to be not that generic

libs/shared/src/components/login/styled.ts
```diff
@@ -0,0 +1,51 @@
...
+export const BasicCard = styled(Card)(({ theme }) => ({
+  display: "flex",
+  justifyContent: "center",
+  width: "100%",
+  maxWidth: theme.spacing(76),
+  marginBottom: theme.spacing(6),
+  padding: theme.spacing(4)
+}));
+
+export const Login = styled(Form)(({ theme }) => ({
```
kacperChlon, line 34: It's BaseLoginForm

libs/shared/src/components/login/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { string } from "yup";
+
+import { REGEX_GMAIL_VALIDATION } from "../../constants";
+import { BaseRoute, FormFieldType, TranslationNamespace } from "../../types";
+import { createPath } from "../../utils";
+import { LocalizedLink } from "../localized-link";
+import * as Styled from "./styled";
+
+export const Login: React.FC = () => {
+  const { i18n, t } = useTranslation(TranslationNamespace.Common);
```
kacperChlon, line 14: do we need this namespace?

libs/shared/src/components/login/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+  const navigate = useNavigate();
+
+  const handleSubmit = (data: unknown) => {
+    navigate(createPath({ route: BaseRoute.Home, language: i18n.language }));
+
+    return console.log(data);
```
kacperChlon, line 20: console log

libs/shared/src/components/login/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+              .matches(REGEX_GMAIL_VALIDATION, t("login.notGmail")),
+            password: string().trim()
+          }}
+          onSubmit={handleSubmit}
+          onError={errors => console.log(errors)}
+          showSubmitButton
+          submitButtonText='Log in with'
+          submitButtonIcon={<Google />}
+        >
+          <LocalizedLink to={BaseRoute.Login}>
```
kacperChlon, line 63: to make it and control position without using children etc... let's add to FormField property appendix: JSX.Element, and display it below given field (and below helper text etc...)

libs/shared/src/constants/regex.ts
```diff
@@ -1 +1,6 @@
+/* eslint-disable no-control-regex */
+
 export const REGEX_URL_FIRST_SLASH = /^\//g;
+
+export const REGEX_GMAIL_VALIDATION =
+  /(?:[a-z0-9!#$%&'*+/=?^_`{|}~-]+(?:\.[a-z0-9!#$%&'*+/=?^_`{|}~-]+)*|"(?:[\x01-\x08\x0b\x0c\x0e-\x1f\x21\x23-\x5b\x5d-\x7f]|\\[\x01-\x09\x0b\x0c\x0e-\x7f])*")@gmail\.com/;
```
kacperChlon, line 6: long one ^^ Have you created it by your own? ;) 

WDYT about:
/([a-zA-Z0-9_\.\+-]{3,}+@gmail\.com)$/

libs/shared/src/utils/create-path.ts
```diff
@@ -4,7 +4,14 @@ import { AppRouteType } from "../types";
 import { buildQueryString } from "./build-query-string";
 import { getAppRoute } from "./get-app-route";
 
-export const createPath = (route: AppRouteType, params?: Params, language?: string, search?: string) => {
+interface CreatePathProps {
```
kacperChlon, line 7: and why exactly you did that? ;) We had typing already provided and right now as you can see we are always casting like (route: SomeRoute, params: RouterParams ...), so why? ;)

libs/shared/src/components/login/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+              .trim()
+              .required(t("login.emailRequired"))
+              .email(t("login.emailInvalid"))
+              .matches(REGEX_GMAIL_VALIDATION, t("login.notGmail")),
+            password: string().trim()
+          }}
+          onSubmit={handleSubmit}
+          onError={errors => console.log(errors)}
+          showSubmitButton
+          submitButtonText='Log in with'
```
kacperChlon, line 60: missing translation

libs/translations/src/common/en.json
```diff
@@ -3,6 +3,15 @@
     "newPresentationDescription": "Placeholder text for ActionCard to create new presentation on Homepage",
     "newPresentationButton": "New presentation"
   },
+  "login": {
+    "title": "Welcome in Patronage App",
+    "subtitle": "Login to your account",
+    "forgotPassword": "Forgot a password?",
+    "emailRequired": "Please provide your e-mail",
+    "emailInvalid": "Invalid e-mail address",
+    "notGmail": "E-mail address must be in Google domain 'example@gmail.com'"
```
kacperChlon, line 12: notGmailMessage

libs/shared/src/components/login/styled.ts
```diff
@@ -0,0 +1,51 @@
...
+  width: "100%",
+  margin: 0,
+
+  "& .MuiFormControl-root": {
```
kacperChlon, line 38: what happen if you change it to root: {...}?

libs/shared/src/components/login/styled.ts
Adys23: Shouldn't we avoid naming components like SomethingSomethingForm ? GoogleLogin?

libs/shared/src/constants/regex.ts
Adys23: Found this one on the internet :)

Yours allows to start with the dot, plus, etc.  :) 

As per gmail website it allows only letters, numbers and dots. So maybe: `/^(\w+[\.\w]{5,}@gmail\.com)$/`

image

libs/shared/src/utils/create-path.ts
Adys23: Without it, if we want to pass only route and string we have to pass undefined for params and language:
`createPath(BaseRoute.Home, undefined, undefined, searchString)`

Thanks to having it like that you can pass only the props you need.

libs/shared/src/components/login/styled.ts
Adys23: Like this?
image

It actually does nothing.

libs/shared/src/components/form-field/index.tsx
Adys23: Sure

libs/shared/src/components/form/index.tsx
Adys23: Deleted

libs/shared/src/components/form/index.tsx
Adys23: Done

libs/shared/src/components/form/index.tsx
```diff
@@ -70,21 +79,25 @@ export const Form: React.FC<FormProps> = ({
...
   return (
     <FormProvider {...methods}>
       <Styled.Form className={className}>
         {title && <Typography variant='h3'>{title}</Typography>}
 
+        {customField}
+
         {renderFields()}
 
+        {children}
```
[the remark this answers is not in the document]
Adys23: Sure, will do it like: https://github.com/intive/patronage22-web-wroclaw/pull/56/files#r820977109

libs/shared/src/components/login/index.tsx
Adys23: We don't. What are the cases where namespace should be provided?

libs/shared/src/components/login/index.tsx
Adys23: Removed

libs/shared/src/components/login/index.tsx
Adys23: Added

libs/shared/src/components/login/index.tsx
Adys23: Done

libs/shared/src/components/login/styled.ts
Adys23: Ok

libs/shared/src/components/login/styled.ts
Adys23: Sure

libs/translations/src/common/en.json
```diff
@@ -3,6 +3,15 @@
     "newPresentationDescription": "Placeholder text for ActionCard to create new presentation on Homepage",
     "newPresentationButton": "New presentation"
   },
+  "login": {
+    "title": "Welcome in Patronage App",
+    "subtitle": "Login to your account",
+    "forgotPassword": "Forgot a password?",
+    "emailRequired": "Please provide your e-mail",
+    "emailInvalid": "Invalid e-mail address",
```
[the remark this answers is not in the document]
Adys23: Ok

libs/translations/src/common/en.json
Adys23: Corrected

libs/shared/src/constants/regex.ts
kacperChlon: works for me ;) My only concern regarding yours was that we could use mails like: 
@gmail.comcomCOM etc... ;) `+` was a bit too long/messy, `/^([a-z0-9]+[a-z0-9.]{4,}[a-z0-9]@gmail\.com)$/` looks good ;)

apps/site/src/app/components/google-login/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+            },
+            {
+              type: FormFieldType.Text,
+              name: "password",
+              defaultValue: "",
+              variant: "filled",
+              description: t("password"),
+              disabled: true,
+              hideEditIcon: true,
+              appendix: <ForgotPasswordLink variant='subtitle2' route={BaseRoute.Home} text={t("login.forgotPassword")} />
```
kacperChlon, line 40: Can we reuse our existing BaseButton component for such scenario? What I mean is instead of creating many different components let's extend (if needed) existing one. Even if here "should" be a text only we can use BaseButton with proper variant :) Sometimes of course it's needed or requested to do so but inn our case let's reuse as manny as possible

apps/site/src/app/components/google-login/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+    <Styled.LoginGoogleContainer>
+      <Styled.LoginGoogleTitle variant='h5'>{t("login.title")}</Styled.LoginGoogleTitle>
+      <Styled.LoginGoogleSubtitle variant='h6'>{t("login.subtitle")}</Styled.LoginGoogleSubtitle>
+      <Styled.LoginGoogleFormCard>
+        <Styled.LoginGoogleForm
+          fields={[
+            {
+              type: FormFieldType.Text,
+              name: "email",
+              defaultValue: "",
```
kacperChlon, line 28: can we use initialValues property for the form to provide initial(default) value for the fields?

kacperChlon: like in the example from my comment/topic

image

kacperChlon: and it'll work like: 

```
...
initialValues: {{ login: "", password: "" }}
...
```

then useForm({
...,
defaultValues: initialValues
})

apps/site/src/app/components/google-login/index.tsx
Adys23: I am already reusing LocalizedLink (I have put it in to separate component to pass it as one line instead of three)

image

apps/site/src/app/components/google-login/index.tsx
Adys23: Why it will be better than having default value set /or not separately for each field?

libs/shared/src/components/login/index.tsx
kacperChlon: For instance in e2e tests where we are not providing default namespaces, or if we'd like to be accurate regarding translationns and for site we'd use by default only common ns or site ns and for other features we'd need to pass it there :) or if we would have main namespace for features like feedback but we'd also have namespace for specific components or abbreviations or timezones etc....

apps/site/src/app/components/google-login/forgot-password-link.tsx
```diff
@@ -0,0 +1,14 @@
+import { Typography, TypographyVariant } from "@mui/material";
+import { AppRouteType, LocalizedLink } from "@patronage-web/shared";
+
+interface ForgotPasswordButtonProps {
```
Reden4005, line 4: Button?

apps/site/src/app/components/google-login/styled.ts
```diff
@@ -0,0 +1,50 @@
...
+  marginBottom: theme.spacing(6),
+  padding: theme.spacing(4)
+}));
+
+export const LoginGoogleForm = styled(Form)(({ theme }) => ({
+  width: "100%",
+  margin: 0,
+
+  "& .MuiFormControl-root": {
+    margin: theme.spacing(0)
```
Reden4005, line 38: Can we use just 0?

apps/site/src/app/components/google-login/forgot-password-link.tsx
Adys23: I don't know... Kacper is saying we should use mer descriptive names. I could probably drop the ...Props ending and leave it as ForgotPasswordButton

apps/site/src/app/components/google-login/styled.ts
Adys23: Sure thing :)

libs/translations/src/common/en.json
```diff
@@ -3,6 +3,16 @@
     "newPresentationDescription": "Placeholder text for ActionCard to create new presentation on Homepage",
     "newPresentationButton": "New presentation"
   },
+  "login": {
```
aniasgit, line 6: translation should be in alphabetical order

libs/translations/src/common/pl.json
```diff
@@ -3,6 +3,16 @@
     "newPresentationDescription": "Tymczasowy tekst komponentu ActionCard do tworzenia nowych ankiet z poziomu widoku Homepage",
     "newPresentationButton": "Nowa ankieta"
   },
+  "login": {
```
aniasgit, line 6: https://github.com/intive/patronage22-web-wroclaw/pull/56/files#r824035344
\

apps/site/src/app/components/google-login/forgot-password-link.tsx
kacperChlon: For sure not button ;)

apps/site/src/app/components/google-login/index.tsx
kacperChlon: because: 
one place where you can set all default values, secondly it's supported by useForm so it should be give us a hint that maybe we can use it ;) secondly we can get rid of defaultProps <- reduce prop driling, and last but not least if your form connfiguration will become more complex it's really handy to have this one place for all initial values for the fields.

apps/site/src/app/components/google-login/forgot-password-link.tsx
```diff
@@ -0,0 +1,14 @@
...
+  variant: TypographyVariant;
+  route: AppRouteType;
+  text: string;
+}
+
+export const ForgotPasswordLink: React.FC<ForgotPasswordButtonProps> = ({ variant, route, text }) => (
```
kacperChlon, line 10: If we've got ...PasswordLink then the props shouldn't be name like ...ButonProps <- I'm not sure am I using Link component or Button component?

Plus based on the component I can tell you this is not ForgotPasswordLink, this is generic componnent that will show me a linked text based onn properties and has nnothing related with ForgotPassword. For instance: 

`<ForgotPasswordLink route="PageRoute.Pricing" variant="primary" text="click to order" />  `

Please channge the name to for instance LinkedText annd move it to shared directory

apps/site/src/app/components/google-login/forgot-password-link.tsx
Adys23: https://github.com/intive/patronage22-web-wroclaw/pull/56#discussion_r824561726 
This is basically what Ola has meant by her comment 
LinkedTextProps is the final version.

libs/translations/src/common/en.json
Adys23: This is very _generic_ :) requirement and from what I see noone was sorting the translations, so this probably should be done for all translation files and mentioned in readme.md

apps/site/src/app/components/google-login/index.tsx
Adys23: Added

apps/site/src/app/components/google-login/forgot-password-link.tsx
Adys23: Done

apps/site/src/app/components/google-login/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { useNavigate } from "react-router-dom";
+import { string } from "yup";
+
+import * as Styled from "./styled";
+
+export const GoogleLogin: React.FC = () => {
+  const { i18n, t } = useTranslation(TranslationNamespace.Common);
+  const navigate = useNavigate();
+
+  const handleSubmit = (_data: unknown, event?: React.BaseSyntheticEvent) => {
```
Reden4005, line 20: Shouldn't we pull BaseSynteticEvent directly from react library?

kacperChlon (approved):

apps/site/src/app/components/google-login/index.tsx
Adys23: We can :)

PR #63, opened 2022-03-14
fix(notifications): move Notifications component to Layout component …

aniasgit:
…in site section


apps/site/src/app/components/layout/index.tsx
```diff
@@ -1,14 +1,29 @@
...
+    ),
+    [FeatureName.Default]: (
+      <DefaultLayout>
+        {children}
+        <Notifications />
+      </DefaultLayout>
+    )
   };
 
   return layouts[useFeatureName()];
```
kacperChlon, line 29:
```
return 
<>
{layouts[useFeatureName()]}
<Notifications />
</>
```

aniasgit: moved to main.tsx

kacperChlon (approved):

PR #64, opened 2022-03-15
JI:760835 P2022-649 Dashboard pagination

pawelwojdas:


libs/shared/src/components/basic-select/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+  name: string;
+  value: ValueType;
+}
+
+export interface BasicSelectProps extends Partial<Pick<SelectProps, "size" | "variant">> {
+  label: string;
+  onChangeHandler: (value: ValueType) => void;
+  items: SelectItem[];
+  value?: ValueType;
+  sx?: SxProps;
```
Adys23, line 16: Why we are using here sx attribute and not the styled-component approach?

libs/shared/src/components/basic-select/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+}
+
+export const BasicSelect: React.FC<BasicSelectProps> = ({ items, onChangeHandler, label, variant, size, value, sx }) => {
+  const handleChange = (event: SelectChangeEvent<ValueType>) => {
+    onChangeHandler(event.target.value);
+  };
+
+  return (
+    <FormControl fullWidth sx={sx}>
```
Adys23, line 25: This looks like generic form use-case :) You might consider adding Select as the new FieldType

libs/shared/src/components/basic-select/index.tsx
pawelwojdas: It's just to overwrite default styles

Reden4005: so maybe name it customStyle.

pawelwojdas: I'm moving this component to our generic form and I'll remove Sx

libs/shared/src/components/form-field/index.tsx
```diff
@@ -2,23 +2,26 @@ import { StandardTextFieldProps } from "@mui/material";
...
+export interface FormFieldProps
+  extends Pick<UseControllerProps, "name" | "defaultValue" | "control">,
+    Partial<Pick<StandardTextFieldProps, "size">> {
   type: FormFieldType;
   variant?: FormTextFieldVariant;
   rows?: number;
   label?: string;
   helperText?: string;
-  onChange?: () => void;
+  onChange?: (event: ChangeEvent<HTMLInputElement>) => void;
```
kacperChlon, line 18: This action shouldn't be extended, it was described during form PR with @Adys23, please contact him to get more explanation

libs/shared/src/components/pagination/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+        count={pagesCount}
+        onChange={handlePageChange}
+        showFirstButton
+        showLastButton
+        size='large'
+        boundaryCount={isMobile ? 0 : 1}
+        siblingCount={isMobile ? 0 : 1}
+      />
+      <Styled.SectionBox>
+        <Styled.InfoBox>{`${itemsPerPageLabel || t("pagination.itemsPerPage")}: `}</Styled.InfoBox>
```
kacperChlon, line 87: please move it to const pageLabel

libs/shared/src/components/pagination/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+  return (
+    <Styled.PaginationBox>
+      <Pagination
+        page={page}
+        count={pagesCount}
+        onChange={handlePageChange}
+        showFirstButton
+        showLastButton
+        size='large'
+        boundaryCount={isMobile ? 0 : 1}
```
kacperChlon, line 83: please move it to const with some descriptive name

libs/shared/src/components/pagination/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+        siblingCount={isMobile ? 0 : 1}
+      />
+      <Styled.SectionBox>
+        <Styled.InfoBox>{`${itemsPerPageLabel || t("pagination.itemsPerPage")}: `}</Styled.InfoBox>
+        <Styled.SizeSelector
+          validationSchema={{}}
+          fields={[
+            {
+              onChange: handleSelectChange,
```
kacperChlon, line 92: to gett value please use onFormChange

pawelwojdas: I couldn't do that, because if I change a value in select it only trigger onClick event, it means that Form component doesn't recognize any change and doesn't trigger that, that's why I call `handleSelectChange` here.

libs/shared/src/components/form-field/index.tsx
pawelwojdas: https://github.com/intive/patronage22-web-wroclaw/pull/64#discussion_r831958260

libs/shared/src/components/pagination/index.tsx
Adys23: handleFormChange function does not need an event at all. It is checking, if the values has changed by its own. Then, as an argument for the onChange function you provide, you get the object with all form values:

`const handleFormChange = () => {
    if (onChange && !isEqual(methods.getValues(), previousValues)) {
      onChange(methods.getValues());
    }
  };`

`onChange?: (value: Record<string, any>) => any;`

libs/shared/src/components/pagination/index.tsx
pawelwojdas: done

libs/shared/src/components/pagination/index.tsx
pawelwojdas: done

libs/shared/src/components/pagination/index.tsx
pawelwojdas: But it doesn't work for onClick event. Only for testing I generated submit button for my select and when I submitted a form it shows the proper value, but still doesn't trigger onChange.

Reden4005 (commented):
LGTM

libs/shared/src/components/pagination/index.tsx
```diff
@@ -0,0 +1,111 @@
+import { Pagination } from "@mui/material";
+import { ChangeEvent, useEffect } from "react";
+import { useTranslation } from "react-i18next";
+
+import { ITEMS_PER_PAGE_OPTIONS } from "../../configs";
+import { useScreenSize, useUrlParams } from "../../hooks";
+import { FormFieldType, PaginationQueryParams } from "../../types";
+import * as Styled from "./style";
```
krysia1, line 8: I think you have to rename it to `styled`

libs/translations/src/common/en.json
```diff
@@ -27,9 +27,14 @@
...
     "title": "We are sorry, the page you're looking for doesn't exist"
   },
+  "pagination": {
+    "itemsPerPage": "Items per page",
+    "outOf": "out of"
+  },
   "password": "Password",
   "save": "Save",
   "search": {
+    "noResultsInfo": "No results found...",
```
krysia1, line 37: It's already added three lines below ;P

libs/shared/src/components/language-button/index.tsx
```diff
@@ -1,26 +1,26 @@
...
-
   const isOtherLang = i18n.language?.length && isOtherLanguage(i18n.language);
-  const langQueryString = isOtherLang ? `?lang=${i18n.language}` : "";
+  const { setParams, removeParams } = useURLParams(["lang"], isOtherLang ? { lang: i18n.language } : null);
 
   const handleChange = (event: React.MouseEvent<HTMLElement>, value: SupportedLanguage) => {
-    changeLanguage(i18n, value);
+    if (value) {
+      changeLanguage(i18n, value);
+      if (!isOtherLang) {
```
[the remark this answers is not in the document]
pawelwojdas: It cannot be stale. Every time when you would use `isOtherLang` it would be a proper value (it's based always on current language). In this case I used exclamation mark, because one line above language was changed with `changeLanguage` function and based on that I've got exactly what I wanted, but it only worked for two languages and it was mistake. If we would add another language it would fail in some cases. For now code is refactored and it works as it should 😄

libs/shared/src/components/pagination/index.tsx
pawelwojdas: Right, I should follow the convention 😄

libs/translations/src/common/en.json
pawelwojdas: My bad... Thanks 👍

libs/shared/src/components/pagination/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const isSizeCorrect = !!sizeFromParam && ITEMS_PER_PAGE_OPTIONS.includes(sizeFromParam);
+  const size = isSizeCorrect ? sizeFromParam : initialSize;
+
+  const pagesCount = Math.ceil(itemsCount / size);
+
+  const isPageCorrect = !!pageFromParam && pageFromParam <= pagesCount && pageFromParam > 0;
+  const page = isPageCorrect ? pageFromParam : INITIAL_PAGE;
+
+  const lastButOnePage = pagesCount - 1;
+  const shownItemsPerPageCount = page === pagesCount ? itemsCount - lastButOnePage * size : size;
```
kacperChlon, line 46: please move calculation to separate operation or move this whole operation to some function

libs/shared/src/components/pagination/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+  const isSizeCorrect = !!sizeFromParam && ITEMS_PER_PAGE_OPTIONS.includes(sizeFromParam);
+  const size = isSizeCorrect ? sizeFromParam : initialSize;
+
+  const pagesCount = Math.ceil(itemsCount / size);
+
+  const isPageCorrect = !!pageFromParam && pageFromParam <= pagesCount && pageFromParam > 0;
+  const page = isPageCorrect ? pageFromParam : INITIAL_PAGE;
+
+  const lastButOnePage = pagesCount - 1;
```
kacperChlon, line 45: any ideas for better name?

libs/shared/src/components/pagination/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const shownItemsPerPageCount = page === pagesCount ? itemsCount - lastButOnePage * size : size;
+  const infoText = `${sizeFromParam >= itemsCount ? itemsCount : shownItemsPerPageCount} ${t("pagination.outOf")} ${itemsCount}`;
+
+  const pageLabel = `${itemsPerPageLabel || t("pagination.itemsPerPage")}: `;
+  const pagesVisibilityCount = isMobile ? 0 : 1;
```
kacperChlon, line 50: can't we do this within pagination component?

libs/shared/src/components/pagination/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+  const itemsPerPageOptions = ITEMS_PER_PAGE_OPTIONS.reduce((acc, option) => {
+    if (itemsCount / option <= ITEMS_PER_PAGE_RATIO && option !== ITEMS_PER_PAGE_OPTIONS[FIRST_ITEM_PER_PAGE_POSITION]) {
```
kacperChlon, line 53: please move math operation out of if scope

libs/shared/src/components/pagination/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+      return acc;
+    }
+
+    return [...acc, { [`${option}`]: option }];
```
kacperChlon, line 57: please use push instead of spread here

libs/shared/src/components/pagination/index.tsx
pawelwojdas: done

libs/shared/src/components/pagination/index.tsx
pawelwojdas: done

libs/shared/src/components/pagination/index.tsx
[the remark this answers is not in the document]
pawelwojdas: done

libs/shared/src/components/pagination/index.tsx
```diff
@@ -0,0 +1,119 @@
+import { Pagination } from "@mui/material";
+import { ChangeEvent, useEffect } from "react";
+import { useTranslation } from "react-i18next";
+
+import { ITEMS_PER_PAGE_OPTIONS } from "../../configs";
+import { useScreenSize, useUrlParams } from "../../hooks";
+import { FormFieldType, PaginationQueryParams } from "../../types";
+import * as Styled from "./styled";
+
+const INITIAL_PAGE = 1;
```
Reden4005, line 10: Maybe will be better to put those constants to one object, name it for example PAGINATION_CONFIG?

libs/shared/src/components/form/form-field/styled.ts
```diff
@@ -1,6 +1,6 @@
 import { Box, styled } from "@mui/material";
```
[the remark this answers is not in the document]
pawelwojdas: right, thanks 😄

kacperChlon (approved):

PR #66, opened 2022-03-17
JI769767 p2022 1771 refactor searchbar

Reden4005:
Replace SearchInput by generic button component.


libs/shared/src/components/base-button/index.tsx
```diff
@@ -7,17 +7,17 @@ export enum ButtonType {
...
 
-export const BaseButton: React.FC<BaseButtonProps> = ({ children, onClick, type, variant, disabled }) => {
+export const BaseButton: React.FC<BaseButtonProps> = ({ children, onClick, type, variant, disabled, startIcon, sx }) => {
   const { isMobile } = useScreenSize();
   const buttonSize = isMobile ? "small" : "medium";
 
   const types = {
     [ButtonType.Basic]: (
-      <Styled.BasicButton size={buttonSize} onClick={onClick} variant={variant} disabled={disabled}>
+      <Styled.BasicButton size={buttonSize} onClick={onClick} variant={variant} disabled={disabled} startIcon={startIcon} sx={sx}>
```
kacperChlon, line 20: In case of adding properties to the generic component please take under consideration that for instance sx prop is also applicable to button type icon.

libs/shared/src/components/search-bar/search-bar.tsx
```diff
@@ -1,8 +1,10 @@
+import SearchIcon from "@mui/icons-material/Search";
+import { t } from "i18next";
```
kacperChlon, line 2: Why like this? We are using useTranslation hook

Reden4005: ok

libs/shared/src/components/base-button/index.tsx
Reden4005: right

libs/shared/src/components/search-bar/search-bar.tsx
```diff
@@ -1,23 +1,34 @@
...
     <>
-      <LocalizedLink to={allResultsPage}>
-        <SearchInput onClick={handleClick} disabled />
-      </LocalizedLink>
+      <Styled.SearchButtonBox>
+        <BaseButton
+          type={ButtonType.Basic}
+          onClick={handleClick}
+          startIcon={<SearchIcon />}
+          disableRipple
```
kacperChlon, line 26: Since we are using ripple in all other places inn application why we would like to disable it here?

libs/shared/src/components/search-bar/search-drawer/index.tsx
```diff
@@ -37,6 +41,18 @@ export const SearchDrawer: React.FC<SearchDrawerProps> = ({ open, onClose, searc
     onClose();
   };
 
+  const handleSubmit = (_data: unknown, event?: BaseSyntheticEvent) => {
```
kacperChlon, line 44: why undescore in data property?

Reden4005: This is implementation / overloading of function, and I have to declare this parameter, however it's not used, so convention required to undersocre it.

Reden4005: I have to declare this parameter, however it's not used, so convention required to underscore it.

libs/shared/src/components/search-bar/search-bar.tsx
Reden4005: It does not fit to this button, looks bad.

libs/shared/src/components/search-bar/search-drawer/index.tsx
Reden4005: I add a convention for unused variables to README file.

kacperChlon (approved):

PR #67, opened 2022-03-19
fix(form): fix onSubmit in form

Reden4005:
Fix onSubmit on Form

### DIFF
--- a/libs/shared/src/components/form/index.tsx
+++ b/libs/shared/src/components/form/index.tsx
@@ -47,7 +47,6 @@ export const Form: React.FC<FormProps> = ({
 
   const methods = useForm<FieldValues>({
     resolver: yupResolver(schema),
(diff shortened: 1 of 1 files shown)


libs/shared/src/components/form/index.tsx
```diff
@@ -47,7 +47,6 @@ export const Form: React.FC<FormProps> = ({
 
   const methods = useForm<FieldValues>({
     resolver: yupResolver(schema),
-    mode: "onChange",
     reValidateMode: "onChange"
```
pawelwojdas, line 50: We can remove `reValidateMode: "onChange"` because it's set in default

Reden4005: Indeed.

kacperChlon (approved):

PR #65, opened 2022-03-17
WIP - JI:769784 P2022-1776 Provide hook for managing query params

pawelwojdas:

PR #69, opened 2022-03-21
JI:769967 P2022-1800 Fix routing

aniasgit:


libs/shared/src/components/language-button/index.tsx
```diff
@@ -18,7 +18,7 @@ export const LanguageButton = () => {
   };
 
   useEffect(() => {
-    setSearchParams(langQueryString);
```
kacperChlon, line 21: I think this won't be valid when this will be merged:
https://github.com/intive/patronage22-web-wroclaw/pull/65

Moreover do you know why you have used replace? I mean what it does?

aniasgit: If I understand correctly without "replace" we have 2 elements in history stack, one with the path before setting searchParams, and the other after. The paths look the same because we are adding searching params when whe are using createPath in LocalizedLink. If "replace = true" setSearchParams replaces the current path in history stack instead of adding new one.

kacperChlon: Yes, it replaces latest item in history, we have to check do we still need to do this after changes :) Please pull latest master :) If it will be needed we have to do this inside useUrlParams hook

kacperChlon (approved):

PR #70, opened 2022-03-21
P2022-1652 E2E tests for Login page & Homepage

Adys23:


apps/site-e2e/src/integration/login-page.spec.ts
```diff
@@ -0,0 +1,27 @@
+describe("Login page", () => {
+  it("opens the login page", () => {
```
Reden4005, line 2: We should confirm that we are on login page, for example: 
cy.get("h5").should("contain", "Welcome in Patronage App")

apps/site-e2e/src/integration/login-page.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+    cy.visit("http://localhost:4000/login");
```
Reden4005, line 3: We should confirm that we are on login page, for example:
cy.get("h5").should("contain", "Welcome in Patronage App")

apps/site-e2e/src/integration/login-page.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+  });
+
+  it("displays e-mail input field and set some value", () => {
+    cy.get("#mui-1").type("Test e-mail").should("have.value", "Test e-mail");
```
Reden4005, line 7: I  am not sure what are you checking here. Maybe will be better to connect this test with "sets correct e-mail in the input" test?

apps/site-e2e/src/integration/login-page.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+  });
+
+  it("checks if error message is shown", () => {
+    cy.get(".Mui-error").should("be.visible");
```
Reden4005, line 11: Does this test pass? If there is no error on the page, why should it be visible?

Reden4005: You should type invalid e-mail first and than check if there is an error.

apps/site-e2e/src/integration/login-page.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+  });
+
+  it("does not redirect after submit with invalid email", () => {
+    cy.get(".MuiButton-root").click();
```
Reden4005, line 15: The tests are independent. It won't work.

apps/site-e2e/src/integration/login-page.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+    cy.url().should("include", "/login");
+  });
+
+  it("sets correct e-mail in the input", () => {
+    cy.get("#mui-1").clear().type("asdasd.asdasd@gmail.com");
```
Reden4005, line 20: This input will be empty, no need to clear it. The email is entered but what is checked? Maybe we can connect it with "redirects to homepage after successful login" test?

apps/site-e2e/src/integration/login-page.spec.ts
```diff
@@ -0,0 +1,19 @@
+describe("Login page", () => {
+  it("opens the login page", () => {
+    cy.visit("http://localhost:4000/login?lang=en");
```
pawelwojdas, line 3: I would add path in env variables, because some values could be different across multiple environments. 
For language we should use enum.

libs/features/feedback-e2e/src/integration/feedback-homepage.spec.ts
```diff
@@ -0,0 +1,17 @@
...
+
+  it("displays welcome text", () => {
+    cy.get(".e1y13pc43 > .MuiTypography-root").contains(
+      "Have you ever wonder how to make your presentation more interactive? Click the button below and simply... make it happen!"
+    );
+  });
+
+  it("redirects to add-presentation page after click", () => {
+    cy.get(".MuiButton-root").click();
+    cy.url().should("include", "/feedback/presentation/add");
```
pawelwojdas, line 15: https://github.com/intive/patronage22-web-wroclaw/pull/70/files#r832039925

libs/features/feedback-e2e/src/integration/feedback-homepage.spec.ts
```diff
@@ -0,0 +1,17 @@
+describe("Feedback Homepage", () => {
+  it("opens the Homepage", () => {
+    cy.visit("http://localhost:4000/?lang=en");
+    cy.url().should("eq", "http://localhost:4000/?lang=en");
+  });
+
+  it("displays welcome text", () => {
+    cy.get(".e1y13pc43 > .MuiTypography-root").contains(
+      "Have you ever wonder how to make your presentation more interactive? Click the button below and simply... make it happen!"
```
pawelwojdas, line 9: We should take this text from translations.  If someone change only one letter - it would fail.

apps/site-e2e/src/integration/login-page.spec.ts
```diff
@@ -0,0 +1,19 @@
...
+    cy.url().should("include", "/login");
+  });
+
+  it("types invalid input and checks if error message is shown and submit is not possible", () => {
+    cy.get("#mui-1").type("Test e-mail").get(".Mui-error").should("be.visible");
```
pawelwojdas, line 8: Shouldn't we set all of these `id's and classes` to variables at the top and then use them?

apps/site-e2e/src/integration/login-page.spec.ts
kacperChlon: not in env variables but in cypress config as baseUrl ;) I think we have this but maybe I forgot to add base url: https://docs.cypress.io/guides/references/configuration#Global

kacperChlon: and after setting this baseUrl we should of course use are function to create proper path

libs/features/feedback-e2e/src/integration/feedback-homepage.spec.ts
kacperChlon: There are 2 ways:
Ideal - we are hardcoding translation to really check whether given translation is correct - for instance if someone will break main translation then we will know about it and we will be sure that it's changed. However every change in translation would require from us to change it in a test and we would need to tak care of all languages. 

Good enough - we can use translation from file which reduces amount of time we have to spend on this however we are not 100% save. 

My solution and let's say compromise between those two ;) 

Create another namespace with translations for e2e - it will be a hard-fork of our site, dashboard, common translations and use this inside our e2e tests.
**Cons:**
- we are multiplying files (not always a bad thing but still can be considered as cons),
- we have 2 sources of truth (not exactly true, since only main namespaces are used in production)
**Pros:** 
- we ensure ourselves we've independent tests 
- we can rely more on our tests
- we still have possibility to test translation for all languages

PR #72, opened 2022-03-31
JI769773 P2022 1775 Logout component

Reden4005:
Goal: As a user I'd like to be able to logout from the app.

AC: 

user should able to logout from admin site
After logout user should see login page


libs/shared/src/components/logout-component/styled.tsx
```diff
@@ -0,0 +1,18 @@
...
+  top: theme.spacing(10),
+  right: theme.spacing(4),
+
+  [theme.breakpoints.down("sm")]: {
+    top: theme.spacing(12),
+    right: theme.spacing(1)
+  }
+}));
+
+export const LoginButtonStyle = {
```
pawelwojdas, line 15: If it's not styled component should it be here?

libs/shared/src/components/logout-component/styled.tsx
```diff
@@ -0,0 +1,18 @@
...
+  fontSize: "16px",
```
pawelwojdas, line 16: Better use values from theme -> e.g. `theme.typography,htmlFontSize`

Reden4005: but it is not styled component, so I can't use theme.

libs/shared/src/components/logout-component/styled.tsx
Reden4005: It is css, so I think it is good place.

libs/shared/src/components/logout-component/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+  const { t, i18n } = useTranslation(TranslationNamespace.Common);
+  const navigate = useNavigate();
+
+  const handleLogout = () => {
+    navigate(createPath({ route: BaseRoute.Login, language: i18n.language }));
+  };
+
+  return (
+    <Styled.LoginButtonWrapper>
+      <BaseButton type={ButtonType.Basic} onClick={handleLogout} sx={Styled.LoginButtonStyle}>
```
Adys23, line 20: why with sx prop? 
Can't it be done in styled.ts like this:
`export const LoginButton = styled(BaseButton)({your styles})`

and then here instead of `BaseButton` use `Styled.LoginButton`

this will allow you to use `theme` as mentioned by @pawelwojdas

libs/shared/src/components/index.ts
```diff
@@ -9,6 +9,7 @@ export * from "./layout";
 export * from "./linked-text";
 export * from "./loader";
 export * from "./localized-link";
+export * from "./logout-component";
```
Adys23, line 12: Why not only "logout"? It is in `src/components` directory, so we can figure out it is a component :)

libs/shared/src/components/logout-component/styled.tsx
```diff
@@ -0,0 +1,18 @@
+import { Box } from "@mui/material";
+import { styled } from "@mui/material/styles";
+
+export const LoginButtonWrapper = styled(Box)(({ theme }) => ({
+  position: "absolute",
```
Adys23, line 5: why do we need absolute positioning?

apps/site/src/app/components/layout/index.tsx
```diff
@@ -8,6 +8,7 @@ export const Layout: React.FC = ({ children }) => {
     [FeatureName.Feedback]: (
       <FeedbackLayout>
         <PreviousPageButton />
+        <LogoutButton />
```
Adys23, line 11: Shouldn't LogoutButton be part of Navbar?

libs/shared/src/components/logout-component/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+  // TODO replace logic when login functionality will be ready
+  const { t, i18n } = useTranslation(TranslationNamespace.Common);
+  const navigate = useNavigate();
+
+  const handleLogout = () => {
+    navigate(createPath({ route: BaseRoute.Login, language: i18n.language }));
+  };
+
+  return (
+    <Styled.LoginButtonWrapper>
```
Adys23, line 19: In `LogoutButton` component we have `LoginButton` elements. I think it is better to stick with Login everywhere.

libs/shared/src/components/logout-component/index.tsx
Reden4005: Will be refatored.

libs/shared/src/components/index.ts
Reden4005: I will rename it

libs/shared/src/components/logout-component/index.tsx
Reden4005: I will change it.

libs/shared/src/components/logout-component/styled.tsx
Reden4005: Changed.

krysia1:
LGTM 👍

libs/shared/src/types/app-routes.ts
```diff
@@ -30,3 +30,12 @@ export const ROUTES: Record<AppRouteType, string | string[]> = {
   [FeedbackRoute.ExternalUserPresentation]: [`/${FeatureName.ExternalFeedback}`, "/presentation", "/:id"],
   [FeedbackRoute.EditPresentation]: [`/${FeatureName.Feedback}`, "/presentation", "/edit/:id"]
 };
+
+export const PROTECTED_ROUTES = [
+  "/",
```
pawelwojdas, line 35: instead using join() use `getAppRoutes()` . For `"/" you can use `BaseRoute.Home`

Reden4005: I think this solution is too complicated. I approached it a little differently.

libs/shared/src/components/navbar/index.tsx
```diff
@@ -20,14 +23,28 @@ export interface NavbarProps extends Pick<AppBarProps, "color" | "position">, Pi
 }
 
 export const Navbar: React.FC<NavbarProps> = ({ config, color, position, variant = "dense" }) => {
+  const location = useLocation();
+  const ProtectedRoute = isProtectedRoute(location);
+
+  const senctionsElements: Record<NavbarSectionPosition, JSX.Element[]> = {
```
pawelwojdas, line 29: there is a typo

libs/shared/src/components/navbar/index.tsx
```diff
@@ -20,14 +23,28 @@ export interface NavbarProps extends Pick<AppBarProps, "color" | "position">, Pi
 }
 
 export const Navbar: React.FC<NavbarProps> = ({ config, color, position, variant = "dense" }) => {
+  const location = useLocation();
+  const ProtectedRoute = isProtectedRoute(location);
```
pawelwojdas, line 27: first letter should be lowercase

libs/shared/src/components/index.ts
```diff
@@ -6,9 +6,11 @@ export * from "./footer";
 export * from "./form";
 export * from "./language-button";
 export * from "./layout";
+export * from "./layout/styled";
```
pawelwojdas, line 9: it's better to change name of `index.tsx` file in layout directory to `layout.tsx` and then create `index.ts` for exporting all of the files in layout. You won't need to export `layout` and `layout/styled` separately.

libs/shared/src/components/logout/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { createPath } from "../../utils";
+import { BaseButton, ButtonType } from "../base-button";
+import * as Styled from "./styled";
+
+export const LogoutButton = () => {
+  // TODO replace logic when login functionality will be ready
+  const { t, i18n } = useTranslation(TranslationNamespace.Common);
+  const navigate = useNavigate();
+
+  const handleLogout = () => {
```
pawelwojdas, line 14: add //TODO logic should be placed in redux when redux-first-router will be applied

libs/shared/src/components/logout/index.tsx
```diff
@@ -0,0 +1,25 @@
+import { useTranslation } from "react-i18next";
+import { useNavigate } from "react-router-dom";
+
+import { BaseRoute, TranslationNamespace } from "../../types";
+import { createPath } from "../../utils";
+import { BaseButton, ButtonType } from "../base-button";
+import * as Styled from "./styled";
+
+export const LogoutButton = () => {
+  // TODO replace logic when login functionality will be ready
```
pawelwojdas, line 10: this comment should placed inside `handleLogout`

libs/shared/src/components/logout/index.tsx
```diff
@@ -0,0 +1,25 @@
+import { useTranslation } from "react-i18next";
```
pawelwojdas, line 1: I would change the name of the file to `logout-button`

libs/shared/src/components/previous-page-button/index.tsx
```diff
@@ -17,7 +17,11 @@ export const PreviousPageButton = () => {
...
   return (
     <Styled.PreviousPageButtonWrapper>
-      <BaseButton type={ButtonType.Icon} onClick={() => navigate(PREVIOUS_PAGE)} sx={Styled.PreviousButtonStyle}>
+      <BaseButton
+        type={ButtonType.Icon}
+        onClick={() => navigate(PREVIOUS_PAGE)}
+        sx={{ fontFamily: "inherit", fontSize: "20px", borderRadius: "8px" }}
+      >
         <Tooltip title={previousPageMessage}>
           <ArrowBackIcon sx={Styled.BackArrowIconStyle} />
```
pawelwojdas, line 26: keep consistency, look at `sx` in `BaseButton`

libs/shared/src/utils/is-protected-route.ts
```diff
@@ -0,0 +1,5 @@
+import { Location, matchPath } from "react-router-dom";
+
+import { PROTECTED_PATHS } from "../types/protected-routes";
```
pawelwojdas, line 3: add `protected-routes` in export file and then `../types` will be sufficient

libs/shared/src/utils/is-protected-route.ts
```diff
@@ -0,0 +1,5 @@
...
+
+export const isProtectedRoute = (location: Location) => PROTECTED_PATHS.find(path => matchPath(path, location.pathname))?.length;
```
pawelwojdas, line 5: this function returns number or undefined so it can't be named like that. You can do something like this -> 
`export const isProtectedRoute = (location: Location) => !!PROTECTED_PATHS.find(path => matchPath(path, location.pathname));`

Reden4005: this function returns boolean

pawelwojdas: image

libs/shared/src/utils/is-protected-route.ts
Reden4005: I tried this. This will create circular dependencies as there is a lot of things initialize as module static. In particular `getAppRoute` can't operate on ROUTES as module is being initialize.

libs/shared/src/components/navbar/index.tsx
Reden4005: thanks

libs/shared/src/components/navbar/index.tsx
Reden4005: yes

libs/shared/src/components/index.ts
Reden4005: ok

libs/shared/src/components/logout/index.tsx
Reden4005: ok

libs/shared/src/components/logout/index.tsx
Reden4005: ok

libs/shared/src/components/logout/index.tsx
Reden4005: ok

libs/shared/src/components/previous-page-button/index.tsx
Reden4005: ok

libs/shared/src/utils/is-protected-route.ts
Reden4005: ok

apps/site/src/app/components/layout/index.tsx
```diff
@@ -2,13 +2,16 @@ import { Layout as FeedbackLayout } from "@patronage-web/features-feedback";
...
 import { Layout as DefaultLayout } from "./default-layout";
+import * as Styled from "./styled";
 
 export const Layout: React.FC = ({ children }) => {
   const layouts = {
     [FeatureName.Feedback]: (
       <FeedbackLayout>
-        <PreviousPageButton />
-        {children}
+        <Styled.LayoutWrapper>
```
krysia1, line 11: Can't we just style "previous page" btn to have `flex-start` styling? Because we want rest of the components to be centered. I checked that this solution should work fine: 
Zrzut ekranu 2022-04-12 o 11 34 17
Zrzut ekranu 2022-04-12 o 11 32 05

Reden4005: Thanks. I will change it.

apps/site/src/app/components/layout/index.tsx
```diff
@@ -1,14 +1,17 @@
...
 
 export const Layout: React.FC = ({ children }) => {
   const layouts = {
     [FeatureName.Feedback]: (
       <FeedbackLayout>
-        <PreviousPageButton />
-        {children}
+        <Styled.PreviousPageWrapper>
+          <PreviousPageButton />
+        </Styled.PreviousPageWrapper>
```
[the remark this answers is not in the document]
Reden4005: I don't think so.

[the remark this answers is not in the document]
piotrrawski: ok, i just thought that we make empty line between elements on the same hierarchy level too

pawelwojdas (commented):
LGTM

libs/shared/src/components/navbar/index.tsx
```diff
@@ -20,14 +23,28 @@ export interface NavbarProps extends Pick<AppBarProps, "color" | "position">, Pi
 }
 
 export const Navbar: React.FC<NavbarProps> = ({ config, color, position, variant = "dense" }) => {
+  const location = useLocation();
+  const protectedRoute = isProtectedRoute(location);
```
kacperChlon, line 27: isProtectedRoute

libs/shared/src/components/navbar/index.tsx
```diff
@@ -20,14 +23,28 @@ export interface NavbarProps extends Pick<AppBarProps, "color" | "position">, Pi
...
   const navSectionElements = Object.values(NavbarSectionPosition).map((section, sectionIndex) => {
     const sectionItem = config[section];
 
     return (
       <Styled.NavbarSectionContainer key={`section-${sectionIndex}`} sx={sectionItem?.customStyles}>
         {sectionItem?.elements.map((element, elementIndex) => (
-          <Styled.NavbarSectionItem key={`element-${elementIndex}`}> {element}</Styled.NavbarSectionItem>
+          <Styled.NavbarSectionItem key={`element-${elementIndex}`}>{element}</Styled.NavbarSectionItem>
         ))}
+        {sectionsElements[section]}
```
kacperChlon, line 47: maybe instead of this logic from 29 to 38 we can just add one func like:

```
const setAdditionalElements = (section: NavbarSectionPosition) => {
  const sectionsElements: Record<NavbarSectionPosition, JSX.Element[]> 
   = {
    [NavbarSectionPosition.Start]: [],
    [NavbarSectionPosition.Center]: [],
    [NavbarSectionPosition.End]: []
  };
  
   sectionsElements[NavbarSectionPosition.End].push(<LogoutButton 
key='logout-button' />);
  }

   return sectionsElements[section]
}
```

libs/shared/src/components/previous-page-button/styled.ts
```diff
@@ -1,23 +1,21 @@
...
 import { styled } from "@mui/material/styles";
 
-export const PreviousButtonStyle = {
-  fontSize: "16px",
-  borderRadius: "8px"
-};
-
-export const BackArrowIconStyle = {
+export const BackArrowIcon = {
   width: "20px",
```
kacperChlon, line 5: let's use theme.spacing

kacperChlon (approved):

libs/shared/src/components/navbar/index.tsx
Reden4005: I can't change it, because I have name collision. I will change it for `isSecuredRoute`.

libs/shared/src/components/previous-page-button/styled.ts
```diff
@@ -1,23 +1,21 @@
...
   display: "block",
 
   [theme.breakpoints.down("md")]: {
     display: "none"
   }
 }));
+
+export const PreviousPageButton = {
+  fontFamily: "inherit",
+  fontSize: "20px",
```
aniasgit, line 19: use theme.spacing here

libs/shared/src/components/navbar/index.tsx
Reden4005: Done

libs/shared/src/components/previous-page-button/styled.ts
Reden4005: ok

libs/shared/src/components/previous-page-button/styled.ts
Reden4005: ok

libs/shared/src/components/previous-page-button/index.tsx
```diff
@@ -14,12 +14,19 @@ export const PreviousPageButton = () => {
...
   const { t } = useTranslation(TranslationNamespace.Common);
   const previousPageMessage = t("goToPage.previous");
+  const theme = useTheme();
 
   return (
     <Styled.PreviousPageButtonWrapper>
-      <BaseButton type={ButtonType.Icon} onClick={() => navigate(PREVIOUS_PAGE)} sx={Styled.PreviousButtonStyle}>
+      <BaseButton
+        type={ButtonType.Icon}
+        onClick={() => navigate(PREVIOUS_PAGE)}
```
pawelwojdas, line 23: instead of trigerring function like this, just create function handle.... and pass it here.

Reden4005: ok

pawelwojdas (approved):

PR #76, opened 2022-04-05
JI773438 P2022 2121 Fix-Incorrect font

Reden4005:
Add Roboto font to our project.


aniasgit (approved):
it fix the problem for me

apps/site/src/index.html
```diff
@@ -4,7 +4,7 @@
     <meta charset="utf-8" name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Patronage Site</title>
     <base href="/" />
-
+    <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto&display=swap" />
```
pawelwojdas, line 7: I would add Helvetica font as well to be consistent with fonts from material UI

Adys23 (approved):
LGTM :)

krysia1:
LGTM 👍

krysia1 (approved):

apps/site/src/index.html
Reden4005: Helvetica is not a free font.

pawelwojdas: Oh, I didn't know

pawelwojdas (approved):

apps/site/src/index.html
kacperChlon: Please use self hosted fonts instead of linked that way. Please provide proper files with proper extensions fonts.

Reden4005: Done

apps/site/src/main.tsx
```diff
@@ -1,4 +1,5 @@
 import "./i18n";
+import "./styles.css";
```
kacperChlon, line 2: change the name to global-styles.css and move to shared directory

apps/site/src/styles.css
```diff
@@ -0,0 +1,44 @@
+/* roboto-italic - latin-ext */
```
kacperChlon, line 1: please move this file to assets or styles folder in shared directory

apps/site/src/styles.css
```diff
@@ -0,0 +1,44 @@
...
+@font-face {
+  font-family: "Roboto";
+  font-style: italic;
+  font-weight: 400;
+  src: url("./fonts/roboto/roboto-italic.eot");
+  src: local(""), url("./fonts/roboto/roboto-italic.eot") format("embedded-opentype"),
```
kacperChlon, line 7: if we don't support local fonts then local() function is not necessary

apps/site/src/main.tsx
Reden4005: Refactored.

apps/site/src/styles.css
Reden4005: ok

PR #85, opened 2022-04-11
JI:773036 P2022-2095 [Visual] [Firefox] Incorrect font of "Previous page" button

krysia1:


libs/shared/src/components/previous-page-button/styled.ts
```diff
@@ -2,8 +2,9 @@ import { Box } from "@mui/material";
 import { styled } from "@mui/material/styles";
 
 export const PreviousButtonStyle = {
-  fontSize: "16px",
-  borderRadius: "8px"
+  fontSize: "1rem",
```
Reden4005, line 5: We don't use rem, it should be in px.

krysia1: for fonts we use rem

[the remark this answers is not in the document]
krysia1: for fonts we use rem

piotrrawski:
LGTM

pawelwojdas (commented):
LGTM

libs/shared/src/components/previous-page-button/styled.ts
```diff
@@ -2,8 +2,9 @@ import { Box } from "@mui/material";
...
+  borderRadius: "8px",
```
kacperChlon, line 6: please use theme.spacing

libs/shared/src/components/previous-page-button/styled.ts
kacperChlon, line 5: Please add this REM value to global styles - for instance btnFontSize

kacperChlon (approved):

